PR #8, opened 2021-01-28
Fix various issues running on OCP

redhatrises:
- Ensure sensor is installed on all nodes including master nodes
- Ensure sensor is privileged so that it can run in OCP

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -19,9 +19,9 @@ daemonset:
   #   necessarily unless running within environments such as OpenShift.
   securityContext:
     runAsUser: 0
-    privileged: false
-    readOnlyRootFilesystem: true
-    allowPrivilegeEscalation: false
+    privileged: true
+    readOnlyRootFilesystem: false
+    allowPrivilegeEscalation: true
 
 image:
   repository: falcon-sensor
--- a/.github/config/ct.yaml
+++ b/.github/config/ct.yaml
@@ -1,2 +1,3 @@
 target-branch: main
 chart-dirs: helm-charts
+validate-maintainers: false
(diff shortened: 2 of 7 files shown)


redhatrises:
Need #7 fixes for the helm chart CI/CD to start working properly.

helm-charts/falcon-sensor/values.yaml
```diff
@@ -19,9 +19,9 @@ daemonset:
   #   necessarily unless running within environments such as OpenShift.
   securityContext:
     runAsUser: 0
-    privileged: false
-    readOnlyRootFilesystem: true
-    allowPrivilegeEscalation: false
+    privileged: true
+    readOnlyRootFilesystem: false
+    allowPrivilegeEscalation: true
```
shawndwells, line 24: Why the updated privileges and readOnly to false?

shawndwells (commented):
Re-ran the linting after merging the CI/CD fix PR.

helm-charts/falcon-sensor/values.yaml
redhatrises: @shawndwells does the Linux kernel sensor require a r/w root filesystem, or no?

redhatrises: @shawndwells the updated privileges are required for the sensor to work correctly on the RHCOS nodes. Otherwise, deployment fails on OpenShift.

redhatrises:
> Re-ran the linting after merging the CI/CD fix PR and it still fails

@shawndwells have to rebase in a PR after a merge otherwise the existing PR branch has outdated config.

Good news is that after a rebase it fails on version bump... so that is good!

shawndwells:
> > Re-ran the linting after merging the CI/CD fix PR and it still fails
> 
> @shawndwells have to rebase in a PR after a merge otherwise the existing PR branch has outdated config.
> 
> Good news is that after a rebase it fails on version bump... so that is good!

Are we ignoring those? If so, ack, feel free to self merge

shawndwells (approved):

redhatrises:
@shawndwells tests are passing now... did a bit more to fix the testing in this PR since testing was enabled in the repo.

PR #10, opened 2021-01-30
Update helm templates to configure and apply deamonset for nodes only

redhatrises:
- This enhancement is allows for future changes to add additional components like the installing both a container sensor and node sensor.

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -19,8 +19,9 @@ node:
     # on the Falcon Sensor working correctly.
     #
     # - User that the container will execute as. Typically necessary to run as root (0).
-    # - Runs the Falcon Sensor containers as privileged containers. Typically not
-    #   necessarily unless running within environments such as OpenShift.
+    # - Runs the Falcon Sensor containers as privileged containers. This is required when
+    #   running the Falcon Linux Sensor on Kubernetes nodes to properly run in the node's
+    #   kernel and to actually protect the node..
     securityContext:
       runAsUser: 0
       privileged: true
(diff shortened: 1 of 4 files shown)


helm-charts/falcon-sensor/values.yaml
```diff
@@ -2,44 +2,48 @@
...
+    labels: {}
+
+    updateStrategy: RollingUpdate
+
+    # Various pod security context settings. Bear in mind that many of these have an impact
+    # on the Falcon Sensor working correctly.
+    #
+    # - User that the container will execute as. Typically necessary to run as root (0).
+    # - Runs the Falcon Sensor containers as privileged containers. Typically not
+    #   necessarily unless running within environments such as OpenShift.
```
shawndwells, line 23: Can we point to documentation, or provide high-level guidance, on when privileged containers would be needed (or not)?

Eg does "such as OpenShift" also include Rancher and Pivotal/VMWare? Where do we send users to learn the answer?

helm-charts/falcon-sensor/values.yaml
```diff
@@ -2,44 +2,48 @@
...
+    securityContext:
+      runAsUser: 0
+      privileged: true
+      readOnlyRootFilesystem: false
+      allowPrivilegeEscalation: true
+
+  image:
+    repository: falcon-sensor
```
shawndwells, line 31: Should the images be renamed to ``sensor-node-container`` to match earlier references in the daemonsets.yaml file?

helm-charts/falcon-sensor/values.yaml
redhatrises: The Falcon Linux sensor (which should be used for nodes) has to be privileged regardless of kubernetes environments.
Updated to state this.

helm-charts/falcon-sensor/values.yaml
redhatrises: Updated them both to be falcon-node-sensor

shawndwells (approved):

PR #34, opened 2021-03-15
Add basic value schema for input validation

redhatrises:


redhatrises:
Looks like gh actions is having problems... please don't merge yet.

redhatrises:
There is a bug in helm that prevents `helm lint` from validating correctly with set values with `values.schema.json`. We will have to wait until helm 3.6 is release before we can use this.

redhatrises:
This is ready for review.

isimluk (approved):
👍

PR #37, opened 2021-03-31
Make tolerations and nodeSelector customizable

uncycler:
This PR allows to customize tolerations and nodeSelector of the DaemonSet.

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -21,7 +21,7 @@ node:
       effect: NoSchedule
 
     nodeSelector:
-      beta.kubernetes.io/os: linux
+      kubernetes.io/os: linux
 
   image:
     repository: falcon-node-sensor
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -15,6 +15,14 @@ node:
 
     updateStrategy: RollingUpdate
 
+    tolerations:
+    # this toleration is to have the daemonset runnable on master nodes
+    - key: node-role.kubernetes.io/master
+      effect: NoSchedule
+
+    nodeSelector:
+      beta.kubernetes.io/os: linux
+
   image:
     repository: falcon-node-sensor
     pullPolicy: Always
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.8.1
+version: 0.9.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.8.1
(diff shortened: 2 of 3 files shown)


redhatrises:
Hello. Thank you for your PR! 

Can you provide more reasoning in this pr for these changes? The tolerations and nodeSelector aren't customizable on purpose. Reasoning is that only linux nodes are supported, and security tools should be deployed across the entire cluster rather than limited to some nodes to prevent cross-lateral movement and attacks.

uncycler:
Regarding tolerations, let say that you have some compute nodes tainted for dedicated usage but still want to have the DaemonSet sheduled on them. Also, with AKS on Azure, Spot instance node pools are tainted.

For nodeSelector, maybe some people would like to have the DaemonSet scheduled only on a specific set of nodes.

shawndwells:
> Regarding tolerations, let say that you have some compute nodes tainted for dedicated usage but still want to have the DaemonSet sheduled on them. Also, with AKS on Azure, Spot instance node pools are tainted.

Did not know about the Azure specifics. 

Ref:
https://docs.microsoft.com/en-us/azure/aks/spot-node-pool

> _A spot node pool will have the label kubernetes.azure.com/scalesetpriority:spot, the taint kubernetes.azure.com/scalesetpriority=spot:NoSchedule, and system pods will have anti-affinity._

> 
> For nodeSelector, maybe some people would like to have the DaemonSet scheduled only on a specific set of nodes.

Perhaps we can enable the capability for customization, and decorate it with warning labels that the 'default crowdstrike way' is to deploy on all nodes?

salanki:
I am about to test deploy CrowdStrike on a 3,000 node cluster. There is absolutely zero chance I will deploy a new DaemonSet to all nodes at once. Being able to set the affinity / nodeSelector is a must, to allow for a phased rollout. There are also plenty of other reasons one might want to disable crowdstrike on certain nodes, ie troubleshooting.

hazcod:
Can we -please- get this merged? No way we are deploying falcon without a proper `nodeSelector`.

helm-charts/falcon-sensor/values.yaml
```diff
@@ -15,6 +15,14 @@ node:
 
     updateStrategy: RollingUpdate
 
+    tolerations:
+    # this toleration is to have the daemonset runnable on master nodes
+    - key: node-role.kubernetes.io/master
+      effect: NoSchedule
+
+    nodeSelector:
+      beta.kubernetes.io/os: linux
```
dschunack, line 24: still depreacted, kubernetes.io/os is the new stable selector

hazcod: FYI @uncycler

dschunack: I think new rebase is need after the release today. We also wait on this to set also "kubernetes.io/arch: amd64".
Looks like no ARM64 images or multiarch image for Linux is available yet.

hazcod:
Can we please get this feature into the helm chart?

domdepasquale:
Does https://github.com/CrowdStrike/falcon-helm/blob/falcon-sensor-1.17.10/helm-charts/falcon-sensor/values.yaml#L19 and/or https://github.com/CrowdStrike/falcon-helm/blob/falcon-sensor-1.17.10/helm-charts/falcon-sensor/values.yaml#L40 help with the functionality you're looking for?

redhatrises:
Closing as this is stale and no response as been provided to the new features.

PR #38, opened 2021-04-02
Enable the Helm Operator

redhatrises:


redhatrises:
Okay... this PR is ready for review

README.md
```diff
@@ -1,6 +1,6 @@
 # CrowdStrike Falcon Helm Chart and Helm Operator
 
-[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/falcon-helm)](https://artifacthub.io/packages/search?repo=falcon-helm)
+[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/falcon-helm)](https://artifacthub.io/packages/search?repo=falcon-helm) [![Docker Repository on Quay](https://quay.io/repository/crowdstrike/falcon-helm-operator/status "Docker Repository on Quay")](https://quay.io/repository/crowdstrike/falcon-helm-operator)
```
shawndwells, line 3:
```suggestion
[![Artifact HUB](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/falcon-helm)](https://artifacthub.io/packages/search?repo=falcon-helm) [![Container Repository on Quay](https://quay.io/repository/crowdstrike/falcon-helm-operator/status "Container Repository on Quay")](https://quay.io/repository/crowdstrike/falcon-helm-operator)
```

shawndwells: ``Docker`` registry or ``container``?

README.md
```diff
@@ -39,6 +41,11 @@ The Falcon Helm chart has been tested to deploy on the following Kubernetes dist
   * Nodes must be Linux distributions supported by CrowdStrike. See [https://falcon.crowdstrike.com/support/documentation/20/falcon-sensor-for-linux#operating-systems](https://falcon.crowdstrike.com/support/documentation/20/falcon-sensor-for-linux#operating-systems) for supported Linux distributions and kernels.
 * Red Hat OpenShift Container Platform 4.6+
 
+The Falcon Helm Operator has been tested to deploy on the following Kubernetes
+distributions:
+
+* Red Hat OpenShift Container Platform 4.6+
```
shawndwells, line 47: Does it make sense to call out the full version number, e.g. ``4.6.20+``? Not sure how backward compatible things are.

README.md
```diff
@@ -49,10 +56,10 @@ The Falcon Helm chart has been tested to deploy on the following Kubernetes dist
 # Deployment Considerations
 
 To ensure a successful deployment, you will want to ensure that:
-1. By default, the Helm Chart installs in the `default` namespace. Best
+1. By default when deploying using a Helm Chart, the Helm Chart installs in the `default` namespace. Best
    practices for deploying to Kubernetes is to create a new namespace.
    This can be done by adding `-n falcon-system --create-namespace` to your
-   `helm install` command.
```
shawndwells, line 55: Should the helm charts create a namespace, eg ``crowdstrike``, if it's best practice not to use ``default``?

README.md
```diff
@@ -49,10 +56,10 @@ The Falcon Helm chart has been tested to deploy on the following Kubernetes dist
...
+   `helm install` command. You do not have to do this when using the Helm Operator.
```
shawndwells, line 62:
```suggestion
   `helm install` command. This is not required when using the Helm Operator.
```

README.md
```diff
@@ -85,13 +92,40 @@ The Helm chart and operator code is in development, and not (yet) supported!
 
 On 24-FEB-2021, CrowdStrike announced (via [press release](https://www.crowdstrike.com/press-releases/advanced-threat-protection-for-cloud-and-container-workloads/)) technology previews are available for deploying Falcon into Azure Kubernetes Service (AKS), Google Kubernetes Engine (GKE), Rancher, and Red Hat OpenShift Container Platform (OCP).
 
-This Helm chart is being developed to automate Falcon sensor deployments into Kubernetes environments. As large-scale testing continues, please note that this Helm chart (and more broadly, sensor deployment into AKS, GKE, Rancher, or OCP) is not yet officially supported nor recommended for production workloads. 
+This Helm operator and chart are being developed to automate Falcon sensor deployments into Kubernetes environments. As large-scale testing continues, please note that this Helm chart (and more broadly, sensor deployment into AKS, GKE, Rancher, or OCP) is not yet officially supported nor recommended for production workloads.
```
shawndwells, line 95:
```suggestion
This Helm operator and chart are being developed to automate Falcon sensor deployments into Kubernetes environments. As large-scale testing continues, please note that this content (and more broadly, sensor deployment into AKS, GKE, Rancher, or OCP) is not yet officially supported nor recommended for production workloads. 
```

README.md
```diff
@@ -85,13 +92,40 @@ The Helm chart and operator code is in development, and not (yet) supported!
...
-The Helm Operator code is under development and not yet documented. It is not recommended to use the Helm Operator.
+
+## Deploy using Make from the Git Repository
+
+1. Deploy the Helm operator
+   ```
+   make deploy IMG=quay.io/crowdstrike/falcon-helm-operator:latest
+   ```
+
+2. Customize the Helm operator deployment by adding your CID and configuring the
```
shawndwells, line 109:
```suggestion
2. Customize the Helm operator deployment by adding your CrowdStrike Customer ID (CID) and configuring the
```

config/manager/manager.yaml
```diff
@@ -31,8 +31,8 @@ spec:
         resources:
           limits:
             cpu: 100m
-            memory: 90Mi
+            memory: 180Mi
           requests:
             cpu: 100m
-            memory: 60Mi
+            memory: 90Mi
```
shawndwells, line 37: Should we document the rational behind these numbers, and how/if/when to adjust them?

config/rbac/role.yaml
```diff
@@ -28,15 +28,61 @@ rules:
...
+# We need to easily pull images regardless of namespace
+- apiGroups:
+  - ""
+  - image.openshift.io
+  resources:
+  - imagestreamimages
+  - imagestreammappings
+  - imagestreams
+  - imagestreamtags
+  - imagetags
```
shawndwells, line 64: Does this mean the helm chart and operator is openshift specific? Do we need to adjust for other distros?

config/samples/crowdstrike_v1alpha1_falconsensor.yaml
```diff
@@ -0,0 +1,36 @@
...
+    billing: null
+    cid: null
+    feature: null
+    memfail_every_n: null
+    memfail_grace_period: null
+    message_log: null
+    provisioning_token: null
+    tags: null
+    trace: null
+    update: null
```
shawndwells, line 22: If these are the default/unchanged values, why copy them?

README.md
redhatrises: You have to purposefully create it through a cli flag with helm. It's not recommended or supported through helm to create a namespace in a template.

shawndwells: It's ok to run `helm --create namespace` but not do it inside a helm template? O_o

Should we have creation of the namespace be a required step earlier in the process?

shawndwells: (meaning certainly believe you, just seems weird!)

config/manager/manager.yaml
redhatrises: The rationale is that the operator was crashing because it was hitting the memory limit. Now, it's not crashing. So if the operator is crashing due to hitting the memory limit, it a bug on our part and a ticket should be opened to fix this because it means a newly fixed build and release.

shawndwells: How did we establish 90Mi is the right initial memory? Under what situations would customers want to tune this up or down? Are there docs we can reference to help customers in this decision?

As an example, if a container has extremely high activity it will generate far more detections/events than a idle web server. At what point should these values be increased? Why are they not the same as the recommended minimum memory for a Linux VM?

config/rbac/role.yaml
redhatrises: At this point, yes, since we need to run in a privileged SCC which is OCP specific. But as we go to other distros, we will look into addressing this.

config/samples/crowdstrike_v1alpha1_falconsensor.yaml
redhatrises: because it is a sample customized CRD... and you might not dig into the helm-chart values. This makes it easier on the user to just modify and run without having to dig deeper.

config/rbac/role.yaml
shawndwells: Should we remove the other k8s distros at https://github.com/CrowdStrike/falcon-helm#kubernetes-cluster-compatability ?

Or make a table about what is supported now vs tentative development roadmap?

| k8s | status |
| OCP | OCP 4.6+ |
| EKS | Planned for May 2020 |
| AKS | Planed for Jul 20202 |

etc

README.md
redhatrises: Mind holding off on this change? I am going to add another badge with another PR for an OLM bundle, and then, I plan on restructuring the readme, which at that point, I plan on fixing this. Feel free to disagree.

shawndwells: No problem!

config/rbac/role.yaml
redhatrises: > Should we remove the other k8s distros at https://github.com/CrowdStrike/falcon-helm#kubernetes-cluster-compatability ?

Those refer to the helm chart and not the helm operator.

> Or make a table about what is supported now vs tentative development roadmap?

Because we have both the helm chart and operator in the same repo and readme, I am going to rework the readme to bring more clarity.

config/manager/manager.yaml
redhatrises: > How did we establish 90Mi is the right initial memory? Under what situations would customers want to tune this up or down? Are there docs we can reference to help customers in this decision?

We established it in our test cluster. Kubernetes will scale the resources up and down to its limit based on what the controller manager needs. Since this is for the operator manager pod, this really shouldn't be configured. If anything, I could probably lower the initial requests and let the cluster scale up/down the memory, and I should probably set a limit for cpu requests.

> As an example, if a container has extremely high activity it will generate far more detections/events than a idle web server. At what point should these values be increased? Why are they not the same as the recommended minimum memory for a Linux VM?

This is for the operator manager pod and not a sensor pod

config/rbac/role.yaml
shawndwells: Spoke to @redhatrises rise. Future PR will clean up the REAMEs to make this more clear.

shawndwells (approved):

PR #39, opened 2021-04-02
Expose the option to set imagePullSecrets on the daemonset

jswoods:
This is required if hosting the falcon image in a private registry.


redhatrises:
Good point! Thanks!

shawndwells:
Where should the documentation around this land?

jswoods:
@shawndwells Good point on the docs. I made https://github.com/CrowdStrike/falcon-helm/pull/40.

PR #47, opened 2021-05-24
Add some of the initial content for Falcon Container

redhatrises:


helm-charts/falcon-sensor/Chart.yaml
```diff
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.9.3
+version: 0.9.900
```
shawndwells, line 18: 0.9.3 --> 0.9.900 ?

helm-charts/falcon-sensor/templates/container_deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+    matchLabels:
+      app: injector
+  template:
+    metadata:
+      labels:
+        app: injector
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1234
```
shawndwells, line 21: would `RUNASUSER > 10000` work?

ref https://kubesec.io/basics/containers-securitycontext-runasuser/

Also ref https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups

helm-charts/falcon-sensor/Chart.yaml
redhatrises: yes... because when falcon container is implemented, I am going to bump version to `1.x`

helm-charts/falcon-sensor/templates/container_deployment.yaml
redhatrises: I'll look into this, as I feel like this can be autogenerated. However, mind creating an issue? Right now, I am trying to get all the content in, and then will do some customization around it.

redhatrises:
I think this should be good now

shawndwells (approved):

PR #50, opened 2021-06-02
Add support for setting serviceAccountName

jmariondev:
Setting `serviceAccountName` is necessary when using PodSecurityPolicies that generally don't allow running privileged pods. In this case, a separate service account must be created and associated with a privileged PodSecurityPolicy.

This adds a small template change to allow setting this value from `.Vaules.node.daemonset.serviceAccountName`.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.9.904
+version: 0.10.0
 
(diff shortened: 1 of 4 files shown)


redhatrises:
Cool! Thanks for this!

I wonder... should we also automate creating a service name as well?

shawndwells:
> Cool! Thanks for this!
> 
> I wonder... should we also automate creating a service name as well?

Probably. But in the famous words of @redhatrises, that'd be outside the scope of this PR ;)

redhatrises:
@jmariondev can you bump the chart version please?

jmariondev:
Bumped to 0.10.0

jmariondev:
It appears the version changed on main after this PR was opened. I've rebased this over the current main.

isimluk (approved):

redhatrises:
Thanks

PR #51, opened 2021-06-04
[helm] configurable tolerations

rustielin:
The daemonset should have configurable tolerations, in addition to being runnable on master nodes.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.9.903
+version: 0.9.904
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.9.903
+appVersion: 0.9.904
 
 
 keywords:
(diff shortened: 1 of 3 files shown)


redhatrises:
Thanks for opening this! Can you provide some more context on what additional configuration tolerations it needs? Thanks!

rustielin:
Hey @redhatrises! Thanks for the quick response.

On my cluster, I have a nodepool with taints. I'd like to get falcon on these nodes as well

redhatrises:
> Hey @redhatrises! Thanks for the quick response.
> 
> On my cluster, I have a nodepool with taints (e.g. `NoExecute`). I'd like to get falcon on these nodes as well

Cool! Thanks! Should we add in the tolerations so that you don't have to customize the deployment?

rustielin:
Tolerations are set by operators themselves, so it's enough to have them as helm values that can be configured. I've tested on my own cluster using my fork of the `falcon-helm` chart, setting my own helm values, and it works as intended.

redhatrises:
> Tolerations are set by operators themselves, so it's enough to have them as helm values that can be configured. I've tested on my own cluster using my fork of the `falcon-helm` chart, setting my own helm values, and it works as intended.

Right. Since you have a bunch of taints, I guess I am seeking input from you on removing the key and effect and only setting
`- operator: "Exists"` which means you wouldn't have to worry about setting tolerations... which from a security perspective is a good way to ensure the sensor gets everywhere.

rustielin:
> seeking input from you on removing the key and effect and only setting

Ah gotcha, understood. I think it's better to leave this up to the operator, for flexibility. In my case I want to run the sensor everywhere, but that might not be the case all the time. Perhaps some documentation regarding the use of tolerations could be useful: e.g. if you want to run the sensor everywhere, use an empty key with operator `Exists`, thereby tolerating everything.

rustielin:
Thanks for the quick turnaround @redhatrises! When can we expect release `0.9.904` to be available in the public helm repo?

redhatrises:
> Thanks for the quick turnaround @redhatrises! When can we expect release `0.9.904` to be available in the public helm repo?

Unfortunately, no. The next release will be the 1.x series because of being in the midst of adding new functionality right now that isn't fully working.

PR #52, opened 2021-06-18
Update Container Sensor configuration

redhatrises:
- Don't allow custom certs for the injector
- Remove runasuser/runasgroup as this breaks on OpenShift
- Update configmap with container entries
- Disable pull secret creation unless configured
- Autogenerate certs for the injector


redhatrises:
Ignore the chart version number failure. It shouldn't be `0.10.0`. This PR also creates the version it should be at.

isimluk (approved):

PR #58, opened 2021-07-20
Add kubeVersion in Chart.yml

redhatrises:
Needed to meet new Rancher requirements. Plus it would be a good thing to have in general.


shawndwells (approved):
Agreed. Helps us self document testing

PR #65, opened 2021-08-13
Update Helm operator

redhatrises:
- Update to operator-sdk 1.10.1
- Namespace should use 'operator' in the name
- Add Falcon Container

### DIFF
--- a/config/manager/controller_manager_config.yaml
+++ b/config/manager/controller_manager_config.yaml
@@ -1,0 +1,10 @@
+apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
+kind: ControllerManagerConfig
+health:
+  healthProbeBindAddress: :8081
+metrics:
+  bindAddress: 127.0.0.1:8080
+
+leaderElection:
+  leaderElect: true
+  resourceName: 811c9dc5.crowdstrike.com
--- a/config/samples/crowdstrike_v1alpha1_falconsensor.yaml
+++ b/config/samples/crowdstrike_v1alpha1_falconsensor.yaml
@@ -4,6 +4,23 @@ metadata:
   name: falcon-helm
 spec:
   # Default values copied from <project_dir>/helm-charts/falcon-sensor/values.yaml
+  container:
+    certExpiration: 3650
+    disableNSInjection: false
+    disablePodInjection: false
+    enabled: false
+    image:
+      pullPolicy: Always
+      pullSecrets:
+        enable: false
+        namespaces: null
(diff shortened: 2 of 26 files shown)


config/manager/controller_manager_config.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
+kind: ControllerManagerConfig
+health:
+  healthProbeBindAddress: :8081
+metrics:
+  bindAddress: 127.0.0.1:8080
```
shawndwells, line 6: Why is one bound to localhost/127 whereas the healthProbe appears to be wild card?

config/manager/controller_manager_config.yaml
```diff
@@ -0,0 +1,10 @@
...
+
+leaderElection:
+  leaderElect: true
+  resourceName: 811c9dc5.crowdstrike.com
```
shawndwells, line 10: Is this a static name? Does it mean something?

config/samples/crowdstrike_v1alpha1_falconsensor.yaml
```diff
@@ -4,6 +4,23 @@ metadata:
   name: falcon-helm
 spec:
   # Default values copied from <project_dir>/helm-charts/falcon-sensor/values.yaml
+  container:
+    certExpiration: 3650
```
shawndwells, line 8: Should this be a variable?

shawndwells: For our NIST friends, for example

redhatrises: This is the variable. It's just the default value for the variable. ;)

config/manager/controller_manager_config.yaml
redhatrises: Since this is operator-sdk created and managed content, documentation is rather sparse on why this is. I believe it is because metrics is only needed to be accessed internally to the cluster whereas the health probe needs to be binded to localhost and the IP given to the pod via the cluster.

config/manager/controller_manager_config.yaml
redhatrises: Yes, this is a static name for the kube controller. Most likely used by the kube-scheduler. It is the name of resource object that is used for locking during leader election and is part of the scaffolding provided by the operator-sdk.

shawndwells:
Got it. Thanks for the descriptions.

PR #69, opened 2021-08-20
Remove sidecar logging for daemonset

redhatrises:
https://github.com/CrowdStrike/Dockerfiles/pull/27 needs to be merged first. But this should effectively remove the need for busybox and a sidecar container to send logs to k8s.


shawndwells (commented):
The original intent of this was for logs to be picked up by k8s. If this is removed, how will logs be shown/gathered?

redhatrises:
> The original intent of this was for logs to be picked up by k8s. If this is removed, how will logs be shown/gathered?

This is being done from the container's entrypoint script now. In other words, it is still being done but in a different way.

PR #85, opened 2021-12-03
Force injector to be restarted after deployment

falcon-pioupiou:
Adding this fix to restart the injector automatically, when the helm chart is played again (adding new namespaces, changing sensor params like tags...)
without this fix, secret TLS certificate is changed but process not restarted so you may have these logs :
http: TLS handshake error from 10.244.0.4:34902: remote error: tls: bad certificate
http: TLS handshake error from 10.244.0.4:34902: remote error: tls: bad certificate

and a delay for Kubernetes to deploy/redeploy new pod (as K8s is waiting an answer for the mutating webhook).

### DIFF
--- a/helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
+++ b/helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
@@ -33,6 +33,8 @@ spec:
       crowdstrike.com/provider: crowdstrike
   template:
     metadata:
+      annotations:
+        rollme: {{ randAlphaNum 5 | quote }}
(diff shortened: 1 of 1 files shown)


redhatrises (changes requested):
Thanks for this. Please make the following updates:
1. This needs to be a configurable option in values.yml to allow those who want to have more control on rolling the injector's deployment. It can be enabled by default.
2. Fix the build failures

redhatrises:
Closing as I believe that there are various aspects of this that have been implemented. Please reopen if not.

Issue #43, opened 2021-04-14
Move CID and provisioning token to a secret


redhatrises:
The CID and provisioning token should probably move to a secret instead of existing in a configmap

redhatrises:
Closing this as ultimately the helm chart will be updated to deploy the operator which is more powerful.

PR #87, opened 2021-12-08
Move CID to Secret Object

cmartintlh:
To fulfill the CID portion of #43, this PR moves the `falcon.cid` value to a secret object created by Helm. 

Additionally, this provides a field `existingSecretName` for a user to leverage an existing Kubernetes Secret for the Deployment/Daemonset to point to. This allows users to leverage other secret management solutions to create the secret object and keep the CID out of source control, where values files may be stored.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.6.0
+version: 1.7.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.6.0
+appVersion: 1.7.0
 
 keywords:
   - CrowdStrike
--- a/helm-charts/falcon-sensor/README.md
+++ b/helm-charts/falcon-sensor/README.md
@@ -41,19 +41,24 @@ helm repo update
 
 The following tables lists the Falcon Sensor configurable parameters and their default values.
 
-| Parameter                   | Description                                          | Default                   |
-|:----------------------------|:-----------------------------------------------------|:------------------------- |
-| `falcon.cid`                | CrowdStrike Customer ID (CID)                        | None       (Required)     |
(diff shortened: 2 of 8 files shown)


redhatrises:
Thanks for opening this PR.

There are a few problems that need to be resolved:

1. falcon.cid needs to remain required. This is intentional and has already tripped up more than a few individuals.
2. Helm charts shouldn't modify objects outside of their control. `existingSecretName` shouldn't be added.
3. The build failures need to be fixed. `oneOf` can't be used.

cmartintlh:
Hi @redhatrises, thanks for your review on this. 

The problem we face at our organization is that we commit our values files to source control (we leverage FluxCD's HelmReleases). Thus, having `falcon.cid` as a required value will expose our Customer ID in Git. The introduction of the `existingSecretName` value would circumvent this issue by allowing the use of a secret created by a tool like [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) or [external-secrets](https://github.com/external-secrets/kubernetes-external-secrets), and then reference the existing secret object in the release. Since we're only referencing an existing secret name for use with the Daemonset/Deployment, the chart would not be modifying the object. 

I'm happy to modify this PR to omit the `existingSecretName` and keep `falcon.cid` as a required value, but curious on your thoughts in the above use case.

redhatrises:
The helm chart will definitely be updated in the next month or so to deploy a CRD with the falcon-operator vs what the helm chart does today. This will allow the sensor to be installed with API credentials outside of the having to specify the CID... which you can still do. However, most of the configuration will now be done via CRD.

cmartintlh:
Thanks for the additional insight @redhatrises. Sounds like it'll be best to wait for that update. Should I close this PR?

redhatrises:
If you want... or keep it open as a reminder as well as for others to use... whatever you prefer.

cmartintlh:
I'll keep it open for now. Thanks!

gazal-k:
The new CRDs mentioned here: https://github.com/CrowdStrike/falcon-operator also don't seem to have any mechanism to set the `client_id` & `client_secret` from a secret.

cookandy:
+1 for allowing us the ability to use existing secrets for customer ID, provisioning tokens, and API tokens.

cmartintlh:
> For anyone experiencing this particular issue, you can use [valuesFrom](https://fluxcd.io/flux/components/helm/helmreleases/#values-overrides) to merge values from a secret into your HelmRelease.

This is a good solution and something I overlooked 😃

redhatrises:
Closing as we won't be supporting secret injection.

Issue #92, opened 2022-01-20
Incorrect type for validating imagePullSecrets


gbucknel:
Hello, 

I'm trying to install the daemonset on EKS with a custom docker registry. 
I'm defining the secret like this : 
```
node:
  image:
    repository: "my-registry.com/falcon-node-sensor"
    pullPolicy: Always
    pullSecrets: my-secret
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
falcon-sensor:
- node.image.pullSecrets: Invalid type. Expected: object, given: string

No matter what format I put in there, it doesn't seem to pass validation. Shouldn't this have 
`-name: <secret name>` ? 

In the actual template it looks like it is just spitting out the secretName, not `-name: my-secret`
https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/daemonset.yaml#L62

Thanks for any help.

gcardonag:
For any readers, we have switched to using the 1.4.0 release of the chart as a temporary workaround while we wait for a release to include the PR with the schema fix.

redhatrises:
This was fixed in 1.6.1 and is set to change in 1.7.0 shortly to be a string so that all that is required is the name of the secret. Closing this issue.

PR #93, opened 2022-01-20
Fixes an issue with the node.image.pullSecrets property

gcardonag:
This is related to https://github.com/CrowdStrike/falcon-helm/issues/92

The property, as defined, expects an object-type value. However, the Pod spec expects an array-type value for the "imagePullSecrets" property. Source: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret

In the current state, if we pass an object to node.image.pullSecrets, we pass schema validation but receive an error from Kubernetes:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(DaemonSet.spec.template.spec.imagePullSecrets): invalid type for io.k8s.api.core.v1.PodSpec.imagePullSecrets: got "map", expected "array"
```

Conversely, if we pass an array, we fail schema validation:
```
coalesce.go:199: warning: cannot overwrite table with non table for pullSecrets (map[])
Error: values don't meet the specifications of the schema(s) in the following chart(s):
falcon-sensor:
- node.image.pullSecrets: Invalid type. Expected: object, given: array
```

Therefore it is impossible to deploy a private image in versions 1.4.1+ of the chart.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.6.0
+version: 1.6.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.6.0
+appVersion: 1.6.1
 
 keywords:
   - CrowdStrike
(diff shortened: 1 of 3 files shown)


mattiasa:
I just ran into the same thing and this solves the issue for me as well. Thanks a lot for the timely fix :)

gcardonag:
To whoever has been approving all the GitHub action pipelines, I apologize for the collection of errors!

redhatrises:
> To whoever has been approving all the GitHub action pipelines, I apologize for the collection of errors!

No worries... That's one of the reasons for CI/CD in the first place!

redhatrises:
thanks for your contribution!

PR #99, opened 2022-03-03
Update readme with sensor support for helm chart range

redhatrises:


redhatrises:
Merging because this is needed for product release.

isimluk (commented):
:+1:

PR #101, opened 2022-03-11
Add ability to pass registry json and some fixes

### LINKED ISSUE #100: Support `registryConfigJSON` for DaemonSet Deployments
Instead of managing this secret separately from the Helm release, add support for specifying a base64 encoded Docker Config to use as an image pull secret.

redhatrises:
- Add some minor fixes around validation and kube version handling for PSP
- Fixes #100

### DIFF
--- a/helm-charts/falcon-sensor/templates/daemonset.yaml
+++ b/helm-charts/falcon-sensor/templates/daemonset.yaml
@@ -57,9 +57,18 @@ spec:
         {{- end }}
         {{- end }}
     spec:
-    {{- if .Values.node.image.pullSecrets }}
+    {{- if and (.Values.node.image.pullSecrets) (.Values.node.image.registryConfigJSON) }}
+      {{- fail "node.image.pullSecrets and node.image.registryConfigJSON cannot be used together." }}
+    {{- else -}}
+    {{- if or (.Values.node.image.pullSecrets) (.Values.node.image.registryConfigJSON) }}
       imagePullSecrets:
+      {{- if .Values.node.image.pullSecrets }}
         - name: {{ .Values.node.image.pullSecrets }}
+      {{- end }}
+      {{- if .Values.node.image.registryConfigJSON }}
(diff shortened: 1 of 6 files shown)


helm-charts/falcon-sensor/templates/daemonset.yaml
```diff
@@ -57,9 +57,18 @@ spec:
         {{- end }}
         {{- end }}
     spec:
-    {{- if .Values.node.image.pullSecrets }}
+    {{- if and (.Values.node.image.pullSecrets) (.Values.node.image.registryConfigJSON) }}
+      {{- fail "node.image.pullSecrets and node.image.registryConfigJSON cannot be used together. Only one of them can be used." }}
```
carlosmmatos, line 61:
```suggestion
      {{- fail "node.image.pullSecrets and node.image.registryConfigJSON cannot be used together." }}
```

helm-charts/falcon-sensor/values.schema.json
```diff
@@ -14,6 +14,10 @@
                     "example": [
```
carlosmmatos, line 14: Does `registryConfigJSON` need to be added to the schema?

redhatrises: Looking into how to best do that because it is multi-object right now (null and string). The schema definitely needs to be updated for more things, but that is a future finalization PR.

helm-charts/falcon-sensor/templates/daemonset.yaml
redhatrises: Fixed

helm-charts/falcon-sensor/values.schema.json
redhatrises: Wasn't sure if helm supported the json schema fully. Looks like it does. Yay! Fixed.

carlosmmatos (approved):
ACK

PR #109, opened 2022-03-29
exclude kube-system and kube-public from webhook config

TomRyan-321:
Ensures resources in kube-system/public namespaces are not targeted by mutatingwebhook and failing if falcon-container injector service is unavailable.


isimluk (approved):

redhatrises:
Ack. Thanks! Per a meeting today, we are going to handle the openshift namespaces differently.

PR #111, opened 2022-03-31
Add lookup functions to handle advanced use cases

### LINKED ISSUE #108: [Request] Support Injection in certain Namespaces or all Namespaces with Authentication
**Request:**
Have a wildcard to target docker config secret in every namespace in cluster, or only allow falcon sensor injection in target namespace

**Problem you're trying to solve. What are you trying to do, and why is it hard?:**
Currently we have lots of namespaces in our cluster (20+ namespaces), and we are using the registry.crowdstrike.com/falcon-container/us-2/release/falcon-sensor Docker image which requires authentication. It is an inconvenience that we have to manually add in each namespace one-by-one into the `container.image.pullSecrets.namespaces` value or else our pods will not run as CS is injected to every pod in our cluster and fails to pull the image.

redhatrises:
- Lookup all namespaces if enabled to add secrets to all namespaces
- Lookup existing certs and use if certification rotation is disabled
- Fixes #108

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.13.0
+version: 1.14.0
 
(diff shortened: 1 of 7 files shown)


isimluk (approved):
:ok_hand: :ok_hand: :ok_hand:

isimluk:
Does it make sense to mention then all-namespaces feature flag in the documentation?

PR #91, opened 2022-01-17
Enable and Allow the customization of serviceAccountName

falcon-pioupiou:
- [Fix] the hardcoded serviceAccountName in the daemonset
- [Fix] the hardcoded serviceAccountName in the clusterrolebinding
- [Fix] the hardcoded serviceAccountName in the SecurityContextConstraints
- [Fix] the hardcoded serviceAccountName in the ServiceAccount
- [Update] the values schema to support the serviceAccountName at root of values
- [Update] the ServiceAccount creation based on values (default : true as the account was created before but not used)
- [Add] the usage of the serviceAccountName for the injector
- [Add] Values : add the custom value for the serviceAccountName

Those changes allows the Falcon Container to be deployed on GKE cluster with Workload Identity (config actually recommended and not yet enforced on GCP side)
https://cloud.google.com/iam/docs/best-practices-for-using-workload-identity-federation

it should also support AWS OIDC as a K8S serviceAccount with annotations (for role binding) is needed

PR #116, opened 2022-04-08
Support service account annotations

gaahrdner:
This commit supports annotations which will allow the Falcon container sensor to utilize both GCP [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to) and (probably) AWS OIDC. https://github.com/CrowdStrike/falcon-helm/pull/91 allowed for the renaming of service accounts however annotations are required.

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -125,8 +125,7 @@ container:
 
 serviceAccount:
   name: crowdstrike-falcon-sa
-  # annotations:
-  #   iam.gke.io/gcp-service-account: my-service-account@my-project.iam.gserviceaccount.com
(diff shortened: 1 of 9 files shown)


helm-charts/falcon-sensor/values.yaml
```diff
@@ -123,7 +123,10 @@ container:
       cpu: 10m
       memory: 20Mi
 
-serviceAccount: crowdstrike-falcon-sa
+serviceAccount:
+  name: crowdstrike-falcon-sa
+  # annotations:
```
redhatrises, line 128: Can you uncomment annotations and make it `annotations: {}`

isimluk (approved):

PR #117, opened 2022-04-13
Fix some logic issues

redhatrises:


helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
```diff
@@ -193,7 +193,9 @@ webhooks:
   - name: {{ $name }}.{{ .Release.Namespace }}.svc
     admissionReviewVersions:
       - v1
+    {{- if lt (int .Capabilities.KubeVersion.Minor) 22 }}
```
isimluk, line 196: People with above average attention to detail (🤷 ) may think that assertion for `KubeVersion.Major` should go along.

isimluk (approved):

helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
redhatrises: Not seeing the value of checking the Major version as there hasn't been a Major version bump, and most likely by the time there is a Major bump, this code probably won't exist anymore.

PR #118, opened 2022-04-18
fix: tolerate taints on control-plane nodes only (by default)

domdepasquale:
There are scenarios, albeit few and far between, where the daemonset pod would get scheduled on nodes that were not supported (or intended to deployed to) e.g. Fargate nodes.

This change also allows for complete control of the tolerations set on the daemonset pods (excluding those set by the K8s API itself)


domdepasquale:
Also, found and applied https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint#renaming-the-node-rolekubernetesiomaster-node-taint as recommended by @redhatrises

PR #123, opened 2022-05-04
Update GH actions

redhatrises:
- Auto-release helm charts on chart version change after merge into main
- Only run when there are changes in `helm-charts` dir

### DIFF
--- a/.github/workflows/helm-release.yml
+++ b/.github/workflows/helm-release.yml
@@ -1,9 +1,11 @@
 name: Release Helm Charts
 
 on:
-  release:
-    types:
-      - created
+  push:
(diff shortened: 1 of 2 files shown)


isimluk (approved):

.github/workflows/helm-release.yml
```diff
@@ -1,9 +1,11 @@
...
 
 on:
-  release:
-    types:
-      - created
+  push:
+    branches:
+      - main
+    paths:
+      - 'helm-charts/**'
```
isimluk, line 8: What about adding `- '.github/**'` here?

PR #128, opened 2022-06-01
fix: add note about namespace label. fixes #114

### LINKED ISSUE #114: falcon-system namespace not being excluded for K8s <1.22
The helm chart expects that the "system" namespaces are annotated with `kubernetes.io/metadata.name`. However, that wasn't the default behavior of the K8s API until version 1.22. It was behind a feature gate at 1.21 https://v1-21.docs.kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name

Therefore, when the falcon-system namespace gets created according to the README.md, it doesn't have any labels or annotations to exclude itself from injection.

We can update the docs to ensure that the namespace has the proper labels and/or mention that 1.22 and beyond will fix this.

E.g. 

For clusters <1.22, or where the `NamespaceDefaultLabelName` feature gate is NOT enabled:

```
kubectl create namespace falcon-system
kubectl label namespace falcon-system sensor.falcon-system.crowdstrike.com/injection=disabled
```

domdepasquale:


lomluca:
I would also remove the if/else logic in these points:
- [container_deployment_webhook.yaml:L210](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml#L210 ) 
- [container_webhook_upgrade.yaml:L48](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/container_webhook_upgrade.yaml#L48 ) 

This logic leads to additionals issues. For example, in EKS you can find the label `kubernetes.io/metadata.name` from 1.21 (the feature is enabled by default) and the chart is not working correctly (label `name` is configured). For EKS < 1.21 you don't find `name` label and you need to add it manually fo existing namespaces, so it's doesn't work automatically (at this point, if you need to add a label in any case, it's better to add that to disable injection).

Therefore, I would keep it simple: just use `kubernetes.io/metadata.name` in the chart, and warns user to use `sensor.falcon-system.crowdstrike.com/injection=disabled` for Kubernetes < 1.21/1.22

PR #148, opened 2022-08-05
fix: docs. fix imagepullpolicy

domdepasquale:
I'm fairly certain message_log is an unusable option.


domdepasquale:
I'll bump the chart version once we verify this fix is necessary

redhatrises:
@domdepasquale should we verify that it exists for all initContainers?

redhatrises (changes requested):
@domdepasquale pretty sure message log is an option in the linux sensor

domdepasquale:
> @domdepasquale pretty sure message log is an option in the linux sensor

I removed the offending commit. thanks @redhatrises

redhatrises (approved):

PR #149, opened 2022-08-09
Improvement: Use ChartApp Version as default for Images

dschunack:
Hi,

We noticed some problems with the current version of your helm chart.

At first, we block in our environment the latest tag due to Security Reasons and Rollback Problems.
The ‘latest’ tag does not actually mean latest, it doesn’t mean anything. It's also not possible at the moment to Rollback to an older Version if anyone noticed problems. In older helm chart versions is also the latest tag used and this mean at the end you Rollback from latest to latest. 

The second problem is a possible version drift between Nodes. If a new version were released and re-tagged with "latest"
it could happen that multiple Versions are running on the different Nodes with out knowing this. You only see that the latest tag/version is used. It's hard to identify the correct version at the end if some node has trouble. 
You need to calculate the sha256 hashes to know which versions is used on the different Nodes.

You describe in your values.yaml that an semantic versioning is used "ChartApp Version", but this is not the Case.
https://github.com/CrowdStrike/falcon-helm/blob/675a957c7acea0d28a86d4e1c55db53804b00f92/helm-charts/falcon-sensor/values.yaml#L40-L41
https://github.com/CrowdStrike/falcon-helm/blob/675a957c7acea0d28a86d4e1c55db53804b00f92/helm-charts/falcon-sensor/values.yaml#L139-L140

You describe also to use a [specific Version to use different Sensor Versions](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/README.md#helm-chart-support-for-falcon-sensor-versions). But, this is at the moment also not working due to the used "latest" Image Tag.

The only thing that is not clear to me at the moment is, which Sensor Version is the current release.
I think it's not the current Chart App Version, or?
https://github.com/CrowdStrike/falcon-helm/blob/675a957c7acea0d28a86d4e1c55db53804b00f92/helm-charts/falcon-sensor/Chart.yaml#L23

You describe in the [Readme.MD](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/README.md#L30-L31) other semantic Versions.

| Helm chart Version | Falcon Sensor Version |
|:-------------------|:----------------------|
| `<= 1.6.x`         | `<= 6.34.x`           |
| `>= 1.7.x`         | `>= 6.35.x`           |

**Feel free to update this PR and to set the correct ChartApp Version that matches to current Sensors Version.**

Last but not least, we switched also Image Pull Policy to 'ifNotPresent'. This reduced the ImagePulls and it prevents to reach the rate limit on docker hub.

Signed-off-by: dschunack <dschunack@web.de>

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.17.6
+version: 1.17.7
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/helm-charts/falcon-sensor/README.md
+++ b/helm-charts/falcon-sensor/README.md
@@ -103,8 +103,8 @@ The following tables lists the more common configurable parameters of the chart 
 |:--------------------------------|:---------------------------------------------------------------------|:--------------------------------------------------------- |
 | `node.enabled`                  | Enable installation on the Kubernetes node                           | `true`                                                    |
 | `node.image.repository`         | Falcon Sensor Node registry/image name                               | `falcon-node-sensor`                                      |
-| `node.image.tag`                | The version of the official image to use                             | `latest`                                                  |
-| `node.image.pullPolicy`         | Policy for updating images                                           | `Always`                                                  |
+| `node.image.tag`                | The version of the official image to use                             | ``                                                        |
+| `node.image.pullPolicy`         | Policy for updating images                                           | `IfNotPresent`                                            |
 | `node.image.pullSecrets`        | Pull secrets for private registry                                    | None       (Conflicts with node.image.registryConfigJSON) |
 | `node.image.registryConfigJSON` | base64 encoded docker config json for the pull secret                | None       (Conflicts with node.image.pullSecrets)        |
 | `falcon.cid`                    | CrowdStrike Customer ID (CID)                                        | None       (Required)                                     |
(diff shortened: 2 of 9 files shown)


redhatrises (changes requested):
Hello,

Thanks for opening this PR. A couple of comments:

1. While you can modify the `values.yaml` file provided by the helm chart, this is not best practice. You should create your own `values.yaml` file (which can be called anything really like `myvalues.yaml`) with the custom values for your environment. Alternatively, you can configure the values using `--set` on the cli with the helm command. When either method is used, the `values.yaml` file provided in the helm chart is overridden with the desired configuration from your environment. Please revert the tag and imagePullPolicy changes in this PR and use a custom yaml file or the `--set` flag. 
2. The chartapp version shouldn't be used for tagging at all. Seems as if the comment needs to be updated rather than adding the chartapp version for the default image tag.
3. The sensor version chart is to provide you a reference to the sensor versions that are supported by the helm chart version and not for sensor tagging.

dschunack:
Hey,

We know that we are able to overwrite the options with our own values.yaml, but this is not why I created this PR.
The most of the common Projects like Grafana,Prometheus did exactly what i wrote and they used the AppVersion in the Chart.yaml as default for the ImageTag if you have a Chart per App. On larger charts with multiple different app containers its defined in the values.yaml with semantic version. Sorry, I don't noticed that you use two different images for the node mode in the Chart. In this case it make no sense to use the Chart AppVersion as ImageTag, but it make absolutely sense to use Semantic Versions instead of the "latest" Image Tag due to the known Reason. It reduced also the Pulls of the Image. 

Short example, on one of our EKS we start 32k PODs per day on 150 Nodes. 

In the Node Mode, you use the Sensor node Image twice for the init and the normal Container. 
With "latest" Tag and Pull "Allways" the Image is 64k pulled per day.

With Semantic Versioning and Pull "IfnotPresent" it's only 150 pulled per Day.

This reduced the API requests, Starting time and also the Bandwidth a lot.

Anyway, my intention was to have a real Helm Package with a stable Chart and App version, but without a ImagePullSecret it's not possible to Pull an Images. CID is not enough here.

redhatrises:
Sounds good. Thanks!

PR #150, opened 2022-08-26
[Feature] Add possibilty to set resources and PriorityClass for Node DaemonSet

dschunack:
Hi,

At the moment it's not possible to set the PriorityClass and Resources for the DaemonSet.
This had some negative effects for the PODs of the DaemonSet.

POD are running in QoS Class "BestEffort" and if a Node-pressure signal occurse the POD will be evicted first.

Without a PriorityClass it's not Guaranteed that a POD will started on the Nodes. Short exmaple: If your reach the POD Limit on smaller Nodes, the POD will not start and stucks in the "Pending" state. With PriorityClass we can Guarantee that the POD started at any time on the Nodes. POD with a lower or without an PriorityClass are automatically evecited to Guarantee the start of the POD.

It looks like this at the moment:

```yaml
% kubectl -n falcon-system get pods -o wide                
NAME                  READY   STATUS    RESTARTS   AGE   IP               NODE                                              NOMINATED NODE   READINESS GATES
falcon-sensor-29l6f   0/1     Pending   0          36m   <none>           <none>                                            <none>           <none>
falcon-sensor-c8mq7   1/1     Running   0          10d   10.194.128.39    ip-10-194-128-39.eu-central-1.compute.internal    <none>           <none>
falcon-sensor-c9fxq   1/1     Running   0          21h   10.194.129.53    ip-10-194-129-53.eu-central-1.compute.internal    <none>           <none>
```
```yaml
kubectl -n falcon-system describe pod falcon-sensor-22shf
Name:         falcon-sensor-22shf
Namespace:    falcon-system
QoS Class:                   BestEffort
```

I added both functionalities for the Node DaemonSet to Guarantee the start of the PODs at any time.

Files Changed in the falcon-sensor Chart:

- Charts.yaml
- values.yaml
- templates/daemonset.yaml
- values.schema.json

Signed-off-by: dschunack <dschunack@web.de>

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.17.7
+version: 1.17.8
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.17.7
+appVersion: 1.17.8
 
 keywords:
   - CrowdStrike
--- a/helm-charts/falcon-sensor/templates/daemonset.yaml
+++ b/helm-charts/falcon-sensor/templates/daemonset.yaml
@@ -114,6 +114,10 @@ spec:
         volumeMounts:
           - name: falconstore
             mountPath: /opt/CrowdStrike/falconstore
+      {{- if .Values.node.daemonset.resources }}
+        resources:
+{{ toYaml .Values.node.daemonset.resources | indent 10 }}
+      {{- end }}
       volumes:
         - name: falconstore-dir
           hostPath:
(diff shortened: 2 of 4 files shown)


redhatrises (changes requested):
Hello,

Thank you for opening this PR. Unfortunately at this time, the setting of resources on the daemonset which is a kernel module is unsupported. It would be best to open a crowdstrike support ticket to address setting limits of a kernel module.

dschunack:
@redhatrises Sorry, but I don't know what do you mean with "this is not supported in the kernel module" ?

[Resource requests](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run) are the standard way to reserve resources on the underlaying node to Guarantee the needed resources between the runtime of the Container by the container runtime and the related cgroup. 

When the kubelet starts a container as part of a Pod, the kubelet passes that container's requests and limits for memory and CPU to the container runtime.  On Linux, the container runtime typically configures kernel [cgroups](https://kubernetes.io/docs/reference/glossary/?all=true#term-cgroup) that apply and enforce the limits you defined.
This functionality is not related to executed Kernel Modules/Libraries of the Container or Application itself. 

This has no impact on the functionality of the Container. Ok, if you set also resource limits this could cause CPU throttling, but this is a different story.

dschunack:
Any update?

617m4rc:
I strongly agree with what has been said already and would like to urge you to accept this PR. Having these Pods without PriorityClass makes the whole solution pointless, as there are many cases where the Pods will be evicted or, maybe even worse, not scheduled in the first place.

redhatrises:
For this PR to be merged, the following changes need to happen:

1. Conflicts need to be resolved
2. The added `resources` configuration must be dropped as CS does not support this
3. The main branch merge commits need to be removed.

dschunack:
@617m4rc Thanks for creating the new PR. I will close my PR after your PR approval.

dschunack:
added with #168

PR #151, opened 2022-08-29
Falcon Container - Allow customization of secret name

falcon-pioupiou:
In some deployment cases, secrets can be provisioned/updated by another method - this update allow to customize Falcon container secret name.

### DIFF
--- a/helm-charts/falcon-sensor/templates/container_secret.yaml
+++ b/helm-charts/falcon-sensor/templates/container_secret.yaml
@@ -4,13 +4,13 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "falcon-sensor.fullname" . }}-pull-secret
+  name: {{ .Values.container.image.pullSecrets.name | default (printf "%s-pull-secret" (include "falcon-sensor.fullname" .)) }}
   namespace: {{ .Release.Namespace }}
 data:
   .dockerconfigjson: {{ $registry }}
 type: kubernetes.io/dockerconfigjson
 {{- if .Values.container.image.pullSecrets.namespaces }}
-{{- $name := (printf "%s-pull-secret" (include "falcon-sensor.fullname" .)) }}
+{{- $name := ( .Values.container.image.pullSecrets.name | default (printf "%s-pull-secret" (include "falcon-sensor.fullname" .))) }}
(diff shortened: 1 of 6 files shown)


redhatrises:
Thanks @falcon-pioupiou 

Just eyeballing this, should the name of the secret that gets created also change, or if the secret name is customized, should `container_secret.yaml` not be created?

falcon-pioupiou:
that's a good point @redhatrises ! nice catch ! I just added the condition to not create secret if secrets are specified by the user.

helm-charts/falcon-sensor/templates/container_secret.yaml
```diff
@@ -1,16 +1,17 @@
 {{- if .Values.container.enabled }}
 {{- if .Values.container.image.pullSecrets.enable }}
+{{- if not .Values.container.image.pullSecrets.name }}
 {{- $registry := .Values.container.image.pullSecrets.registryConfigJSON }}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "falcon-sensor.fullname" . }}-pull-secret
+  name: {{ .Values.container.image.pullSecrets.name | default (printf "%s-pull-secret" (include "falcon-sensor.fullname" .)) }}
```
falcon-pioupiou, line 8: I've kept the secret name changed here to keep the consistency (even if not displayed) in case conditions will be changed

helm-charts/falcon-sensor/templates/container_secret.yaml
```diff
@@ -1,16 +1,17 @@
...
   namespace: {{ .Release.Namespace }}
 data:
   .dockerconfigjson: {{ $registry }}
 type: kubernetes.io/dockerconfigjson
 {{- if .Values.container.image.pullSecrets.namespaces }}
-{{- $name := (printf "%s-pull-secret" (include "falcon-sensor.fullname" .)) }}
+{{- $name := ( .Values.container.image.pullSecrets.name | default (printf "%s-pull-secret" (include "falcon-sensor.fullname" .))) }}
```
falcon-pioupiou, line 14: I've kept the secret name changed here to keep the consistency (even if not displayed) in case conditions will be changed

redhatrises:
Thanks much!

PR #153, opened 2022-08-31
Modify function check in node psp template to enable PSP on v1.24 clusters

r3motecontrol:
Hey Team,

PSP's are being deprecated in v1.25, but they are still supported in v1.24. Please modify the second argument to 25 to enable PSP on v1.24 clusters, otherwise the deployment fails.

Cheers


redhatrises:
Awesome. Thanks!

PR #155, opened 2022-08-31
Add nodeAffinity. Fix a few values

domdepasquale:
This adds nodeAffinity for the DS as an optional configuration. Allowing one to set additional node selections e.g. processor type:
```yaml
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
            - amd64
```

Also, I removed an unusable falconctl and moved nameOverride and fullnameOverride to the root of the values since it was nested inside of the daemonset options and not used there.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.17.9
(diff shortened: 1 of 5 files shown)


redhatrises:
Looks good. Should we go further and make it easier so that only:
```
- key: kubernetes.io/arch
            operator: In
            values:
            - amd64
```

domdepasquale:
I left it flexible in case someone wanted to use `preferredDuringSchedulingIgnoredDuringExecution ` ... though I'm not sure that would ever be used in this scenario

PR #168, opened 2022-10-15
[Feature] Add possibilty to the PriorityClass for the DaemonSet

617m4rc:
Presently, setting the PriorityClass for the DaemonSet Pods is impossible. Pods may be evicted or not be scheduled at all in case of resource shortage on the worker nodes.

This PR enables the possibility of setting a PriorityClass, such as system-node-critical, via the Pod template and is an updated version of #150, excluding the aspects of resource requests and limits.


redhatrises:
Please bump the Chart version

redhatrises:
Ack

PR #174, opened 2022-11-07
Fix helm for resources in container agent; add helm for resources to node agent

### LINKED ISSUE #173: Enable resource requests/limits on daemonset pods
Please enable me to set a resource request/limit on the daemonset pods.  

The structure already exists around the container injection variant, and I'd like the same on DS pods so that my cluster autoscaler can correctly understand how much space needs to be provisioned to fit one of the falcon pods.

When karpenter (our cluster autoscaler) detects unschedulable workloads, it provisions a new node to fit those workloads on it.

For example, if I have 2Gi of memory requests that the scheduler cannot fulfill, karpenter provisions an instance that provides (2Gi + daemonset requests) of memory. If the falcon sensor does not participate in this calculation, my nodes are undersized by whatever amount of resource utilization the falcon sensor would require, and either a daemonset pod is unschedulable (very bad) or a workload pod is unschedulable, causing me to create another instance which is then subject to the same problem again (also bad, but differently).

So I and others need to be able to specify the resource requests for all pods, daemonset or otherwise.

jcogilvie:
Fixes #173 

Adds resource support to `.Values.node.daemonset.resources`.

Technically the old version of the container agent would support resourcing if you passed `.Values.resources` but the values file has the resources under `.Values.container.resources` (which would not have worked had someone used it).  Moving those resources to the intended place is technically a breaking change, so bumping the version to 1.18 as well.


redhatrises:
Thank you for opening this PR. We do not support setting resouces at all on the daemonset because of the very nature 
that it is a kernel module. See https://github.com/CrowdStrike/falcon-helm/discussions/146

jcogilvie:
@redhatrises I will.  This seems like a misunderstanding of the nature of resource requests.  

The resource requests are for the scheduler to make room so the pod always has space to run.  It's more about setting the minimum than the maximum.

redhatrises:
> @redhatrises I will. This seems like a misunderstanding of the nature of resource requests.
> 
> The resource requests are for the scheduler to make room so the pod always has space to run. It's more about setting the minimum than the maximum.
> 
> When karpenter (our cluster autoscaler) detects unschedulable workloads, it provisions a new node to fit those workloads on it.
> 
> For example, if I have 2Gi of memory requests that the scheduler cannot fulfill, karpenter provisions an instance that provides (2Gi + daemonset requests) of memory. If the falcon sensor does not participate in this calculation, my nodes are undersized by whatever amount of resource utilization the falcon sensor would require, and either a daemonset pod is unschedulable (very bad) or a workload pod is unschedulable, causing me to create another instance which is then subject to the same problem again (also bad, but differently).
> 
> So I and others need to be able to specify the resource `requests` for _all pods_, daemonset or otherwise.

If anything at this point in time, limits won't be supported. `requests` could be supported. It just needs to be supported officially.

jcogilvie:
@redhatrises that would be completely acceptable, thank you.

redhatrises:
> @redhatrises that would be completely acceptable, thank you.

@jcogilvie make sure to make the request for the daemonset sensor to provide the minimum cpu/mem requirements to run as a pod in kubernetes and not around helm support. Once those requirements have been provided, helm support can be added through a pull request like this one.

jcogilvie:
I can back that data out of the running cluster, but I can add it to a
support case if it's helpful.

It's unconventional for a helm chart to be opinionated about resource
requirements given the wide variety of cluster contexts one can find
oneself in. That's why the convention is to just let me, the operator,
handle it.

redhatrises:
> I can back that data out of the running cluster, but I can add it to a support case if it's helpful. It's unconventional for a helm chart to be opinionated about resource requirements given the wide variety of cluster contexts one can find oneself in. That's why the convention is to just let me, the operator, handle it.

This is where the misunderstanding is. It isn't a helm chart problem at all. It is a sensor problem that has to be addressed at the sensor level BEFORE it can be enabled in the helm chart.

jcogilvie:
Can you elaborate?  

What problems would be caused by saying, "in the past week the kubernetes-reported memory usage for the falcon-sensor container has not grown beyond 20MiB and the cpu usage has held constant at 200-300mcores, so provision those going forward"?  (Just to take the reported metrics used from one of these node agents.)

Most daemonset pods that I'm aware of, including very important and deeply-buried ones like CNI plugins, don't require introspective knowledge of their own memory configuration.  Is there something that makes this one different?

redhatrises:
Closing this until this is supported. The falcon sensor is a kernel module which is buried deeper than a CNI plugin.

gazal-k:
It shouldn't matter how "deep it is buried". I believe we could add support for resource requests, even if the kernel module or the eBPF mode sensor don't honour it. The point would be to allow compute provisioners (like https://karpenter.sh/) to calculate the capacity requirements somewhat more accurately at the least.

Also, it looks like there's only the appearance of this being a repository that expects community contributions. If they're not welcome, this helm chart may as well be closed source.

jcogilvie:
Here's the ID of the idea board submission:  IDEA-I-9590

I second what @gazal-k said; I don't really care about kernel module memory/cpu.  I'm no expert on k8s kernel modules, but it appears there's already a built in [reserved capacity](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/) for that.  So the heart of my request is still as it was:  let me tell Karpenter how much *Node Allocatable* (to use the term from the link) space this pod needs.

Kulagin-G:
I agree with @jcogilvie, this simple DS improvement is required for proper resource management for cluster-autoscaler or karpenter. I'm ok with no limits, but requests are really needed.

fradee:
Any updates for resources for daemonset mode for helm chart v1.19.x?

PR #175, opened 2022-11-07
Feat/injector annotations

domdepasquale:
I have a [scenario](https://cloud.google.com/kubernetes-engine/docs/troubleshooting/troubleshooting-security#fails) where I need to add pod annotations to the injector. This PR does that and also allows additional labels and annotations to be added to the injector Deployment.


redhatrises:
ack

PR #177, opened 2022-12-02
Feat/backend

domdepasquale:


helm-charts/falcon-sensor/values.yaml
```diff
@@ -196,6 +196,7 @@ falcon:
   app:
   trace: none
   feature:
+  backend:
```
redhatrises, line 199: Let's go ahead and set `kernel` here.

helm-charts/falcon-sensor/Chart.yaml
```diff
@@ -15,12 +15,12 @@ type: application
...
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.17.18
+version: 1.17.19
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.17.18
+appVersion: 1.17.19
```
redhatrises, line 23: Since this requires a specific sensor, I wonder if we should bump to `1.18`?

domdepasquale: Incoming change

helm-charts/falcon-sensor/values.yaml
domdepasquale: Incoming change

domdepasquale:
Setting to DRAFT as I don't want it merged quite yet

redhatrises:
Ack

PR #178, opened 2022-12-02
fix: permissions and tests fixes/updates

redhatrises:
- Don't use a custom privileged SCC. Use the default OCP one.
- Set initContainer securitycontext
- Add option to explicitly enable tests so that resources don't appear on clusters


helm-charts/falcon-sensor/templates/clusterrole.yaml
```diff
@@ -35,7 +35,9 @@ rules:
   - securitycontextconstraints
   resourceNames:
 {{- if .Values.node.enabled }}
-  - {{ include "falcon-sensor.fullname" . }}-node
+  - privileged
+  verbs:
+  - use
```
domdepasquale, line 40: When I tested this out, I got some double verbs:

```
rules:
- apiGroups:
  - ""
  - image.openshift.io
  resources:
  - imagestreams/layers
  verbs:
  - get
- apiGroups:
  - security.openshift.io
  resources:
  - securitycontextconstraints
  resourceNames:
  - privileged
  verbs:
  - use
  verbs:
  - use
```

a "dry-run=server" apply didn't fail though

domdepasquale: `helm template --validate --values=ci/cid-values.yaml --set node.enabled=true .`

PR #185, opened 2023-02-10
added pod topology for injector replicas and updated readme

rbojja-cs:


helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
```diff
@@ -84,6 +84,17 @@ spec:
     {{- end }}
     spec:
       affinity:
+        podAntiAffinity:
```
jsullivan-CS, line 87: How would you feel about making these affinity statements default values, and allowing for them to be overridden?

helm-charts/falcon-sensor/values.yaml
```diff
@@ -81,7 +81,37 @@ container:
   enabled: false
 
   # Configure the number of replicas for the mutating webhook backend
-  replicas: 1
+  replicas: 2
+
+  affinity:
+    # Setting Pod anti-affinity soft or preferred rule so that replicas will try to run on different nodes
+    podAntiAffinity:
```
pflanagan-cs, line 88: Just mentioning `topologySpreadConstraints` in case they haven't been considered.
The same result can be achieved with something like
```
topologySpreadConstraints:
- maxSkew: 1
  topologyKey: kubernetes.io/hostname
  whenUnsatisfiable: ScheduleAnyway
  labelSelector:
    matchLabels:
      crowdstrike.com/component: crowdstrike-falcon-injector
```
topologySpreadConstraints are a little more flexible than anti-affinity rules, there is a good explanation [here](https://kubernetes.io/blog/2020/05/introducing-podtopologyspread/)

rbojja-cs: Implemented it using topologySpreadConstraints after confirming about supported K8s version

helm-charts/falcon-sensor/values.yaml
```diff
@@ -81,7 +81,33 @@ container:
...
+            operator: In
+            values:
+            - linux
+     # Setting node affinity rule so that replicas will try to run on worker nodes
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 1
+        preference:
+          matchExpressions:
+          - key: node-role.kubernetes.io/master
+            operator: DoesNotExist
```
redhatrises, line 110: Why was this change made and put into values?

rbojja-cs: @redhatrises I was doing this change when I tried to add podAntiAffinity to affinity statements and provide user to override the default affinity values.But since I am using pod topology spread and no longer updating affinity statements, I just left this as it is in values.yaml.Is there any reason to not put in values.yaml? I can revert back this change if you think there is no reason to allow affinity statements.

redhatrises:
@rbojja-cs can you remove commit [42a2555](https://github.com/CrowdStrike/falcon-helm/pull/185/commits/42a25553c1864fa8453b9dae2e50e21a4fd4325b) ? These type of commits muddy the history which makes reviewing past history problematic.

redhatrises (approved):

PR #188, opened 2023-02-16
breaking fix: move falcon.backend to node.backend

jsullivan-CS:
Relates to https://github.com/CrowdStrike/falcon-operator/pull/332

The current behavior of this chart is to generate the FALCONCTL_OPT_BACKEND environment variable (via configmap) when falcon.backend is set (as it is, by default), regardless of whether the node or container sensor is being deployed.
This variable and corresponding command argument causes a fatal error in version 6.51+ of the container sensor.


falcon-pioupiou (approved):
Good to me

PR #189, opened 2023-02-17
Enable setting maxUnavailable as percentage

gdziwoki:
This PR enables setting maxUnavailable as percentage.
[DaemonSetSpec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec) clearly defines the  maxUnavailable field as IntOrString

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.18.5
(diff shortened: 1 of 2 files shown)


redhatrises:
Thanks for this. The daemonset template needs to also be updated for this to be merged.

gdziwoki:
@redhatrises what needs to change in the template?

redhatrises:
> @redhatrises what needs to change in the template?

Please make sure that this logic works with these changes https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/daemonset.yaml#L35

gdziwoki:
That check was done before opening the PR. All should work fine.

redhatrises:
@gdziwoki can you bump the version to pass CI please so that this can be merged?

PR #192, opened 2023-03-02
Allowing vault value path

abelnieva:


redhatrises:
CID needs to be validated as it is one of the most common missed items... so if you can have the vault item validated then this would be a win.

redhatrises:
Closing this as there is no further movement.

ohamadeh:
> so if you can have the vault item validated then this would be a win.

But there are so many other secret injection solutions that need to be validated as well. I don't think its worth. @abelnieva I think you can use `--skip-schema-validation` option when doing helm install/upgrade

PR #195, opened 2023-03-29
fix: update docs for arm64 support

redhatrises:


CS-kyleweishaar:
Can you add a note about multiarchitecture clusters are not supported.

Requires a x86_64 or ARM64 Kubernetes cluster (multi-architecture clusters are not supported)

PR #200, opened 2023-06-01
Allow flexible configuration of container affinity

### LINKED ISSUE #134: Feature: Add option to limit the Container Sensor to Fargate Pods
Currently, if an Amazon EKS Cluster contains both worker nodes and Fargate nodes/pods, the only way to achieve full coverage of the cluster using the falcon-sensor is to inject the Container Sensor as a sidecar to all pods.

Instead, we would like to be able to deploy both the node sensor and the container sensor as part of the same helm installation. This can be achieved by limiting the container sensor to those Fargate pods by adding a toleration to the falcon-sensor-injector Deployment spec. For example:

```
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: kubernetes.io/os
                operator: In
                values:
                - linux
              - key: eks.amazonaws.com/compute-type
                operator: In
                values:
                - fargate
```

617m4rc:
Allow more flexible configuration of the container affinity. This change allows for better support of mixed scenarios where the regular worker nodes are protected by the node sensor and only certain containers should be protected by the container sensor, e.g. when using AWS Fargate, Azure Container Instances, Google Cloud Run. Resolves #134

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.19.1
+version: 1.19.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/helm-charts/falcon-sensor/README.md
+++ b/helm-charts/falcon-sensor/README.md
@@ -5,7 +5,7 @@ platform purpose-built to stop breaches via a unified set of cloud-delivered
 technologies that prevent all types of attacks — including malware and much
 more.
 
-# Kubernetes Cluster Compatability
+# Kubernetes Cluster Compatibility
 
 The Falcon Helm chart has been tested to deploy on the following Kubernetes distributions:
 
--- a/helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
+++ b/helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
@@ -83,21 +83,10 @@ spec:
         {{- end }}
     {{- end }}
     spec:
+      {{- if .Values.container.affinity }}
       affinity:
-        nodeAffinity:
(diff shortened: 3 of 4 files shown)


redhatrises:
Hello,

Thanks for opening this, but unfortunately, this PR doesn't work as described or intended since it only sets the affinity for the webhook injector service and not the actual sidecar injected sensor. The container sensor uses a mutating webhook which is cluster-wide so setting the affinity for the injection service won't actually tell the sensor to only inject into pods on say fargate-only nodes. This means that in mixed clusters the sensor will be injected into all pods. 

Setting `sensor.falcon-system.crowdstrike.com/injection` to `disabled` in each namespace/pod or disabling injection by default and setting `sensor.falcon-system.crowdstrike.com/injection` to `enabled` in each namespace/pod is the way to do it which is pretty standard/common for kubernetes deployments.

bellaned:
> Hello,
> 
> Thanks for opening this, but unfortunately, this PR doesn't work as described or intended since it only sets the affinity for the webhook injector service and not the actual sidecar injected sensor. The container sensor uses a mutating webhook which is cluster-wide so setting the affinity for the injection service won't actually tell the sensor to only inject into pods on say fargate-only nodes. This means that in mixed clusters the sensor will be injected into all pods.
> 
> Setting `sensor.falcon-system.crowdstrike.com/injection` to `disabled` in each namespace/pod or disabling injection by default and setting `sensor.falcon-system.crowdstrike.com/injection` to `enabled` in each namespace/pod is the way to do it which is pretty standard/common for kubernetes deployments.

Hi,

Is it actually possible to run falcon helm chart with:
**node:
  enabled: true
container:
  enabled: true**

Because in our setup we have issue with using both node sensors and container sensors - using argo for the deployment. 
And as I've reached the falcon support they said to me that this is not possible - to have both sensors run simultaneously.

Thanks in advance!

redhatrises:
> bellaned

Yes. However, this would only be for mixed cluster like EKS with Fargate nodes. Which in that case, you would need to disable the sidecar injection in each namespace or pod not running on Fargate nodes. What is not supported is running both the sidecar in pods that run on nodes that have the sensor also installed.

PR #208, opened 2023-07-05
kweishaar: add readme for KPA docs

CS-kyleweishaar:
@cs-pvyas 
Please review. Our console docs guide users through the console set up after cloud account registration so we'll probably want to adjust the install agent section. I included some of the extra info around the agent: proxy and resource consumption.


CS-kyleweishaar:
Version bumped to 0.43.9

PR #212, opened 2023-08-02
fix: Falcon Linux Sensor image is mistakenly used in the container sensor injector 

617m4rc:
Fix the situation where an apparently wrong image is used for the sensor injector. This situation can occur in mixed scenarios where parts of the cluster are protected using the Falcon Linux sensor while other parts are protected using the Falcon container sensors.

Use only the container sensor image to avoid conflicts with the enforced PodSecurityPolicies, in particular error "container has runAsNonRoot and image has non-numeric user".

Values for reproduction:

```
node:
  enabled: true
  image:
    repository: xxx.dkr.ecr.eu-central-1.amazonaws.com/falcon-node-sensor

container:
  enabled: true
  image:
    repository: xxx.dkr.ecr.eu-central-1.amazonaws.com/falcon-container-sensor
```


helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -63,6 +63,15 @@ Create the name of the service account to use
...
 {{- define "falcon-sensor.image" -}}
 {{- if .Values.node.enabled -}}
+{{ include "falcon-sensor-node.image" . }}
+{{- else -}}
+{{ include "falcon-sensor-container.image" . }}
+{{- end -}}
+{{- end -}}
+
+
+{{- define "falcon-sensor-node.image" -}}
```
redhatrises, line 73: Shouldn't lines 64-72 now be dropped and daemonset/node instances of falcon-sensor.image be changed to "falcon-sensor-node.image"?

helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
```diff
@@ -142,7 +142,7 @@ spec:
       {{- end }}
       containers:
       - name: {{ include "falcon-sensor.name" . }}-injector
-        image: "{{ include "falcon-sensor.image" . }}"
+        image: "{{ include "falcon-sensor-container.image" . }}"
```
redhatrises, line 145: Make sure the InitContainer has this as well.

redhatrises (changes requested):

redhatrises:
Hello,

Thank you for your PR. Unfortunately, this still won't fix the entirety of the problem at all since an image name/tag can be anything. Not opposed to these changes, so a couple of things for merging:

1. Drop the `Merge branch 'main' of https://github.com/CrowdStrike/falcon-helm` commit `2d396c1` as we don't want to merge in and dirty up the git history
2. The commit author (y9011) in this case really should be the PR contributor (617m4rc)

617m4rc:
This tool is too expensive to fix bugs myself.

PR #216, opened 2023-08-23
Fix: sensor: APD ConfigMap value is missing when false

GarrettPoore:
Creating the `falcon-sensor` with the following configuration mangles the proxy settings:
```yaml
falcon:
  cid: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA-AA
  apd: false
  aph: hostname
  app: 1234
```

Resulting ConfigMap values:
```yaml
data:
  FALCONCTL_OPT_CID: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA-AA
  FALCONCTL_OPT_APH: "hostname"
  FALCONCTL_OPT_APP: "1234"
  FALCONCTL_OPT_TRACE: "none"
  FALCONCTL_OPT_BACKEND: "kernel"
```

After this fix:
```yaml
data:
  FALCONCTL_OPT_CID: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA-AA
  FALCONCTL_OPT_APD: "false"
  FALCONCTL_OPT_APH: "hostname"
  FALCONCTL_OPT_APP: "1234"
  FALCONCTL_OPT_TRACE: "none"
  FALCONCTL_OPT_BACKEND: "kernel"
```

APD needs to be "false" when enabling the proxy (disable = false, so enable), so this value is _only_ needed when it's false.

The admission controller helm chart has similar configurations, but the proxy settings are configured as strings, so this boolean check is fine there.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.20.2
+version: 1.21.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.20.2
+appVersion: 1.21.0
 
 keywords:
   - CrowdStrike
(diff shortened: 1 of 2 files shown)


redhatrises:
@GarrettPoore can you please bump the patch instead of the minor version? So `1.20.3` instead of `1.21.0`. Thanks!

GarrettPoore:
> @GarrettPoore can you please bump the patch instead of the minor version? So `1.20.3` instead of `1.21.0`. Thanks!

Sure thing, I assumed that this fell into "minor functionality changes" more than "documentation or typo changes".

GarrettPoore:
@redhatrises Can you approve the workflow? I ran the `ct lint` command locally and I think the latest change should pass.

GarrettPoore:
@redhatrises Is anything else needed for this PR to get it merged in? This is blocking our testing of the sensor in our cluster. I can work around it not being in the Helm chart, but I'd like to avoid the extra work if this is going to be merged in soon.

redhatrises:
Ack. Thanks!

PR #223, opened 2023-09-30
Update README.md for OpenShift

shawndwells:
Adds openshift to the platforms


redhatrises:
@shawndwells Thanks for opening this PR! However OpenShift is no longer supported as the certified operator is to be used instead.

redhatrises:
Closing with the above.

PR #228, opened 2023-10-17
Remove AGENT_CLIENT_ID from the KPA configmap

r3motecontrol:
The `AGENT_CLIENT_ID` and `AGENT_CLIENT_SECRET` are used to create a secret object, which in turn are set as environmental variables for the container by referencing the secret. There is no need to set the `AGENT_CLIENT_ID` environmental variable by adding it again to the configmap.


redhatrises:
Ack. Thanks!

Issue #215, opened 2023-08-18
falcon-sensor chart does not allow seamless upgrade from Kubernetes 1.24 to 1.25


jcmcken:
I think the Helm chart for falcon-sensor does not appropriately deal with Kubernetes upgrades. Consider ``PodSecurityPolicy`` (PSP), which has been deprecated for many versions and is removed entirely in Kubernetes 1.25. Imagine you are at Kubernetes 1.24 with this chart installed. The chart forces you to install the PSP even though PSP, as an API object, is removed in 1.25. Many Kubernetes platforms don't let you upgrade if unsupported object types (e.g. PSPs) are present in the cluster.

Thus there's a chicken-and-egg problem. You need to be at 1.25 for this Helm chart to cease installing the PSP. But if you already use this Helm chart, you can't upgrade to 1.25 safely. Thus you need manual action to remove the PSP. This isn't a scalable approach for enterprises with many clusters, where the installation of tooling is automated and managed, and in many cases operators do not even have direct login access to clusters.

Ideally, every template in the chart that relies on Helm capabilities to detect whether something should be done should also have some value override users can set to simulate that behavior. Otherwise you are opening up the risk of hitting upgrade-related issues without recourse (aside from forking the chart, which is also not ideal).

PR #229, opened 2023-10-20
fix: allow updating in place from k8s 1.24 to 1.25 by optionally disabling PSPs

congiv:
fixes: #215 

Allow the node and container PodSecurityPolicy manifests to be disabled.

If a user has already installed falcon-sensor on a cluster on 1.24, they can use these values to disable the PSPs _prior_ to attempting the upgrade to 1.25. This allows the 1.25 upgrade to proceed cleanly without having to manually patch the PSPs out of the Chart's release secret. 

image


redhatrises:
Hello. Thank you for your PR! Fortunately/Unfortunately, we have just released version 1.21.1 which removes PSP support completely which means this PR no longer works. Going to close this PR since 1.21.1 was released. But, thanks again!

Issue #230, opened 2023-10-23
Allow the use of preexisting PriorityClasses with falcon-sensor


congiv:
With CrowdStrike falcon-sensor 1.21.1 we are no longer able to use preexisting PriorityClasses on the falcon-sensor DaemonSet.

Attempting to do so results in error messages like:
> Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: PriorityClass "my-priority-class" in namespace "" exists and cannot be imported into the current release

This prevents the use of any preexisting PriorityClass, such as [the well-known ones](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical).

It would be nice if this chart allowed us to use any arbitrary PriorityClass defined outside of this chart as it did prior to 1.21.1.

PR #231, opened 2023-10-23
fix(falcon-sensor): allow for preexisting PriorityClass

congiv:
fixes: #230 

Allows for the user to specify a preexisting PriorityClass by opting out of the creation of one in this Chart.

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -24,6 +24,7 @@ node:
     labels: {}
 
     # Assign a PriorityClassName to pods if set
+    priorityClassCreate: true #Whether or not to create a PriorityClass as part of this Chart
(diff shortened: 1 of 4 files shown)


helm-charts/falcon-sensor/values.yaml
```diff
@@ -24,6 +24,7 @@ node:
     labels: {}
 
     # Assign a PriorityClassName to pods if set
+    priorityClassCreate: true #Whether or not to create a PriorityClass as part of this Chart
```
redhatrises, line 27:
```suggestion
    # Enable the priorityClass creation on chart installation
    priorityClassCreate: true
```

redhatrises (changes requested):

redhatrises (approved):

PR #232, opened 2023-10-25
KPA: Apply clientID validation logic only when existingSecret is empty

r3motecontrol:
Apply clientID validation logic only when existingSecret is empty. We are using an `existingSecret` and not hardcoding the `clientID` and `clientSecret` in `values.yaml`. Right now, `clientID` validation is performed even if the `existingSecret` is defined. As a result, this is breaking our deployment.


redhatrises:
LGTM

PR #236, opened 2023-11-16
Feat/kpa proxy support

TomRyan-321:
This PR for the KPA chart adds the following:

- Enables setting the HTTP_PROXY, HTTPS_PROXY & NO_PROXY settings at install time
- Increments chart version
- Resets app version back to .0 increment as this version represents when new images are available and should not have been incremented during the previous merges

### DIFF
--- a/helm-charts/cs-k8s-protection-agent/templates/configmap.yaml
+++ b/helm-charts/cs-k8s-protection-agent/templates/configmap.yaml
@@ -9,14 +9,7 @@ data:
   AGENT_DEBUG: {{ .Values.crowdstrikeConfig.enableDebug | default "false" | quote }}
   AGENT_ENV: {{ .Values.crowdstrikeConfig.env | quote }}
   AGENT_HELM_VERSION: {{ .Chart.Version | quote }}
-  {{- if .Values.proxyConfig }}
-  {{- if .Values.proxyConfig.HTTP_PROXY }}
-  HTTP_PROXY: {{ .Values.proxyConfig.HTTP_PROXY }}
-  {{- end }}
-  {{- if .Values.proxyConfig.HTTPS_PROXY }}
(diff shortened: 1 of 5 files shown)


helm-charts/cs-k8s-protection-agent/values.yaml
```diff
@@ -51,3 +51,5 @@ crowdstrikeConfig:
   cid: ""
   dockerAPIToken: ""
   existingSecret: ""
+
+proxyConfig: {}
```
redhatrises, line 55: Please add the proxy options here as well

helm-charts/cs-k8s-protection-agent/templates/configmap.yaml
```diff
@@ -9,3 +9,14 @@ data:
   AGENT_DEBUG: {{ .Values.crowdstrikeConfig.enableDebug | default "false" | quote }}
   AGENT_ENV: {{ .Values.crowdstrikeConfig.env | quote }}
   AGENT_HELM_VERSION: {{ .Chart.Version | quote }}
+  {{- if .Values.proxyConfig }}
```
redhatrises, line 12: This can be dropped once the proxy options are also added to the default values.yaml file

TomRyan-321: If people are upgrading charts and using deploying using `--reuse-values` its probably still safer to have the if statements as this is not required configuration. Thoughts?

helm-charts/cs-k8s-protection-agent/values.yaml
TomRyan-321: Added

helm-charts/cs-k8s-protection-agent/templates/configmap.yaml
TomRyan-321: Remove condition and set defaults

Issue #241, opened 2023-12-04
falcon-kac: hardcoded pods in ResourceQuota blocking falcon-kac deployment


dttung2905:
Hi team,

In the README doc, we have the example deployment in `falcon-kac` and this works perfectly fine.
```
  helm upgrade --install falcon-kac $KAC_HELM_REPO \
     -n falcon-kac --create-namespace \
     --set falcon.cid=$FALCON_CID \
     --set image.repository=$KAC_IMAGE_REPO \
     --set image.tag=$KAC_IMAGE_TAG
```

However, the problem arises when we deploy in other namespaces, for example `kube-system`. The hard limit of pods is hardcoded in ResourceQuota object which might prevent the new kac deployment to be spun up. The reason is because there are existing deployment with priority class `system-cluster-critical` like the API server, scheduler, and controller-manager
https://github.com/CrowdStrike/falcon-helm/blob/654dd90b111f06e9fd69b73545567937e1e5adac/helm-charts/falcon-kac/templates/resourcequota.yaml#L8-L16

The proposed solution would be to parameterise this value `spec.hard.pods` to values.yaml file and set a default value of 2 to maintain backward compatibility

redhatrises:
Hello,

A couple of things.

1. The admission controller should only be deployed in its own namespace and not in namespaces with existing resources. Deploying in a namespace with existing resources is not a best practice.
2. There should not be multiple installs of the admission controller in the cluster.

PR #242, opened 2023-12-04
Parameterise the hard limit on pods in ResourceQuota object

dttung2905:
Hi team,

I want to parameterise `spec.hard.pods` in `ResourceQuota` object. This is useful when falcon-kac is deployed on to namespace where there are existing deployments that take up pods quota with `PriorityClass=system-cluster-critical` ( for example: api-server in kube-system namespace

Fixes: https://github.com/CrowdStrike/falcon-helm/issues/241

### DIFF
--- a/helm-charts/falcon-kac/templates/resourcequota.yaml
+++ b/helm-charts/falcon-kac/templates/resourcequota.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "falcon-kac.labels" . | nindent 4 }}
 spec:
   hard:
-    pods: 2
+    pods: {{ .Values.resourceQuota.pods }}
   scopeSelector:
     matchExpressions:
     - operator: In
--- a/helm-charts/falcon-kac/values.schema.json
+++ b/helm-charts/falcon-kac/values.schema.json
@@ -259,6 +259,32 @@
                     ]
                 }
             }
+        },
+        "resourceQuota": {
+            "type": "object",
+            "properties": {
+                "pods": {
+                    "oneOf": [
+                        {
+                            "type": [
+                                "string",
+                                "null"
+                            ],
+                            "default": "2",
+                            "pattern": "^[0-9]+$"
+                        },
+                        {
+                            "type": [
+                                "integer",
+                                "null"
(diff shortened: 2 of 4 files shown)


helm-charts/falcon-kac/templates/resourcequota.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     {{- include "falcon-kac.labels" . | nindent 4 }}
 spec:
   hard:
-    pods: 2
+    pods: {{ .Values.resourceQuota.pods | default 2 }}
```
redhatrises, line 10:
```suggestion
    pods: {{ .Values.resourceQuota.pods }}
```

helm-charts/falcon-kac/values.schema.json
```diff
@@ -259,6 +259,16 @@
...
                 }
             }
+        },
+        "resourceQuota": {
+            "type": "object",
+            "properties": {
+                "pods": {
+                    "type": "integer",
+                    "default": "2",
+                    "minimum": 2
```
redhatrises, line 269: This can be an integer or a string and should be handled for either use-case.

dttung2905: Thanks for the review. Are you saying something like? 
```yaml
# values.schema.json file
        "resourceQuota": {
            "type": "object",
            "properties": {
                "pods": {
                    "oneOf": [
                        {
                            "type": "integer",
                            "default": "2",
                            "minimum": 2
                        },
                        {
                            "type": "string",
                            "default": "2"
                        }
                    ]
                }
            }
        }

# resourcequota.yaml
spec:
  hard:
    pods: {{ .Values.resourceQuota.pods | quote }}
```

redhatrises: Sorry, I missed this question, but yes, something like that were there is the user of oneOf.

dttung2905:
Hi @redhatrises and @pflanagan-cs , is there anything that needs to be done for this PR to get merged?

pflanagan-cs (approved):
LGTM

PR #245, opened 2023-12-20
add read permissions for resources that are monitored by KAC

pflanagan-cs:
For dry-run mode, falcon-kac needs read permissions on the resources it validates as admission controller webhook

### DIFF
--- a/helm-charts/falcon-kac/templates/clusterrole.yaml
+++ b/helm-charts/falcon-kac/templates/clusterrole.yaml
@@ -10,6 +10,9 @@ rules:
   resources:
   - namespaces
   - nodes
+  - pods
+  - replicationcontrollers
+  - services
   verbs:
   - get
   - list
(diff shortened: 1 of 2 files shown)


helm-charts/falcon-kac/templates/clusterrole.yaml
```diff
@@ -19,6 +22,25 @@ rules:
...
+  - batch
+  resources:
+  - cronjobs
+  - jobs
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - "admissionregistration.k8s.io"
```
redhatrises, line 41:
```suggestion
  - admissionregistration.k8s.io
```

redhatrises:
To be clear, we aren't talking about helm's `dry-run` correct?

helm-charts/falcon-kac/templates/clusterrole.yaml
pflanagan-cs: fixed

pflanagan-cs:
> To be clear, we aren't talking about helm's `dry-run` correct?
Correct. This is to allow a dryrun of KAC admission controller to see what errors it would report if it was reviewing existing resources

PR #247, opened 2024-01-04
falcon-helm: Update initContainer args

r3motecontrol:
The arguments in the initContainer check if the executable file `/opt/CrowdStrike/falcon-daemonset-init` exists. Since the hostPath volume `falconstore-dir` is mounted to `/host_opt`, this file is not visible inside the initContainer, and the check always fails. I assume this is the helper binary to initialize `falconstore`?

The mountPath change from `/opt` to `host_opt` was introduced in this [commit](https://github.com/CrowdStrike/falcon-helm/commit/1fff4b642babf31000f5240da003cb193c07192c)

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.24.1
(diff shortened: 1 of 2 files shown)


redhatrises:
Hello,

This change is actually incorrect and introduces a bug. `/opt/CrowdStrike/falcon-daemonset-init` exists in `/opt` in the initContainer and `/host_opt` is the node's `/opt`. `/opt/CrowdStrike/falcon-daemonset-init` does not exist in the node's `/opt`. Please make sure that you are using the latest sensor, and if needed,

PR #248, opened 2024-01-09
:sparkles: (feat): Addded ability to reference existing pullsecret

KrisJohnstone:
Added the ability to reference pullsecrets via linking secret to the service account.

### DIFF
--- a/helm-charts/cs-k8s-protection-agent/templates/serviceaccount.yaml
+++ b/helm-charts/cs-k8s-protection-agent/templates/serviceaccount.yaml
@@ -11,3 +11,7 @@ metadata:
   annotations:
   {{- toYaml . | nindent 4 }}
   {{- end }}
+{{- with .Values.image.existingPullSecret }}
+imagePullSecrets:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


redhatrises:
Hello,

Can you provide some additional context on why this method is needed?

KrisJohnstone:
> Hello,
> 
> Can you provide some additional context on why this method is needed?

When using GitOps to deploy this chart, the only provision provided for pullsecrets at present is by committing the pullsecret to git which is bad juju.

redhatrises:
> > Hello,
> > Can you provide some additional context on why this method is needed?
> 
> When using GitOps to deploy this chart, the only provision provided for pullsecrets at present is by committing the pullsecret to git which is bad juju.

Does setting `node.image.pullSecrets` to the existing pull secret not work?

KrisJohnstone:
> > > Hello,
> > > Can you provide some additional context on why this method is needed?
> > 
> > 
> > When using GitOps to deploy this chart, the only provision provided for pullsecrets at present is by committing the pullsecret to git which is bad juju.
> 
> Does setting `node.image.pullSecrets` to the existing pull secret not work?

That requires me to commit a pull secret to git unencrypted.

redhatrises:
> That requires me to commit a pull secret to git unencrypted.

This is incorrect. Setting `node.image.pullSecrets` to the name of your custom secret is what you are wanting and attempting to do, e.g.:

```
--set node.image.pullSecrets=mysecretname
```
Please use this to not store your secret in git unencrypted.

KrisJohnstone:
Apologies was in a rush and must have been confused with the line underneath then which talks about base64.

PR #256, opened 2024-01-26
cleanup: remove priorityclass deployment capability

redhatrises:
- Priorityclass should be passed to the helm chart instead of created by the chart to prevent unintentional user side affects. GKE autopilot has also changed the way how it handles resources which is what this functionality was originally intended for. This makes PC easier to handle.


redhatrises:
Closing for now as there are sadly othe GKE oddisms around priorityclass and resourcequota. May just need a refactor.

PR #257, opened 2024-01-31
fix(kpa): update regsecret's name when registryConfigJSON is used

falcon-pioupiou:
in deployment.yaml line 32 : the secret name should end with `-registryconfigjson`

```
- name: {{ include "cs-k8s-protection-agent.fullname" . }}-regsecret-registryconfigjson
```

This is fixing the secret name

### DIFF
--- a/helm-charts/cs-k8s-protection-agent/templates/docker-secret.yaml
+++ b/helm-charts/cs-k8s-protection-agent/templates/docker-secret.yaml
@@ -1,4 +1,4 @@
-{{ if .Values.crowdstrikeConfig.dockerAPIToken }}
+{{ if or (.Values.crowdstrikeConfig.dockerAPIToken) (.Values.image.registryConfigJSON) }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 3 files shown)


helm-charts/cs-k8s-protection-agent/templates/docker-secret.yaml
```diff
@@ -15,7 +15,7 @@ data:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "cs-k8s-protection-agent.fullname" . }}-regsecret
```
redhatrises, line 18: Looking at the code, the name shouldn't change. What needs to change is the `docker-secret` yaml, and the deployment yaml should be condensed since they basically do the same thing. I can do it in another PR if you don't have the time to make the changes.

redhatrises (changes requested):

helm-charts/cs-k8s-protection-agent/templates/docker-secret.yaml
falcon-pioupiou: I've condensed the secret (to use a condition) depending the value configured, and on deployment side I've used the same checks as Falcon-sensor and falcon-kac to fail the template if `crowdstrikeConfig.dockerAPIToken` is used with `image.registryConfigJSON`

redhatrises:
Ack

PR #261, opened 2024-02-08
feat(iar): adding registryConfigJSON and fixing the pullSecret usage

falcon-pioupiou:
- pullSecret property is not working : it was creating a secret based on the value and not doing a reference to it
- adding the support of the registryConfigJSON to have a consistent way to deploy like other agents (simpler to use than the dockerAPIToken based on the tooling we have today)


helm-charts/falcon-image-analyzer/templates/docker-secret.yaml
```diff
@@ -8,16 +8,10 @@ metadata:
...
-  labels:
-  {{- include "falcon-image-analyzer.labels" . | nindent 4 }}
-type: kubernetes.io/dockerconfigjson
-data:
-  .dockerconfigjson: {{ .Values.image.pullSecret }}
-  {{- end}}
+{{- if .Values.image.registryConfigJSON }}
+  .dockerconfigjson: {{ .Values.image.registryConfigJSON }}
+{{- end }}
+{{- end }}
```
cs-pvyas, line 17: Add new Line

helm-charts/falcon-image-analyzer/values.yaml
```diff
@@ -31,6 +31,12 @@ image:
   # private repository
```
cs-pvyas, line 31: Add a comment here that `pullSecret` should be a reference to a secret in the same namespace - `falcon-image-anlyzer`

PR #264, opened 2024-02-12
Default priorityClassCreate to false

sfotony:
If a user removes `priorityClassCreate` from the values file entirely, the default action should not be to attempt to create a priorityClass of a given `priorityClassName`.  

In my opinion/experience, if the user is using the default helm values file as a guide and has indicated the name of a priority-class to use but has deleted the directive to create a priority class by that name, the intended action is to *not* try to create that priority-class.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.25.2
+version: 1.25.3
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.25.2
+appVersion: 1.25.3
 
 keywords:
   - CrowdStrike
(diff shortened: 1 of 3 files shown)


sfotony:
> Please make sure to make this change in `values.yaml` as well and bump the chart minor version.

I bumped the chart, but I don't know if this necessarily needs to be made in values.yaml.  If example value is `true` that's fine by me, I was more concerned with the implied default being `true`.  If it's the wish of the maintainers to have the values file match the defaults, I'm happy to do that, however.

redhatrises:
> > Please make sure to make this change in `values.yaml` as well and bump the chart minor version.
> 
> I bumped the chart, but I don't know if this necessarily needs to be made in values.yaml. If example value is `true` that's fine by me, I was more concerned with the implied default being `true`. If it's the wish of the maintainers to have the values file match the defaults, I'm happy to do that, however.

Please make the change in `values.yaml` as well as it is the default values configuration (not example) for the helm chart. `values.schema.json` is for only schema validation and doesn't set the default chart values.

sfotony:
> > > Please make sure to make this change in `values.yaml` as well and bump the chart minor version.
> > 
> > 
> > I bumped the chart, but I don't know if this necessarily needs to be made in values.yaml. If example value is `true` that's fine by me, I was more concerned with the implied default being `true`. If it's the wish of the maintainers to have the values file match the defaults, I'm happy to do that, however.
> 
> Please make the change in `values.yaml` as well as it is the default values configuration (not example) for the helm chart. `values.schema.json` is for only schema validation and doesn't set the default chart values.

Ok, updated as requested!

PR #265, opened 2024-02-15
KAC-feat add clusterName support

falcon-pioupiou:
Adding the cluster name support to be displayed in the Host management when the KAC agent is not able to get the clustername. Note: if the agent is able to get a cluster name this value (clusterName) will be overwritten by the agent.


helm-charts/falcon-kac/values.yaml
```diff
@@ -56,6 +56,11 @@ falcon:
   tags:
   provisioning_token:
 
+# When clusterName is not present (like Rancher or microk8s), you can setup the hostname
```
pflanagan-cs, line 59:
```suggestion
# Falcon KAC can usually discover the clusterName automatically so setting clusterName here is normally 
# not necessary, but in some cases the clusterName cannot be discovered by the KAC, e.g. for self-hosted 
# kubernetes clusters such as MicroK8s. In those case, the clusterName can be manually set here.
# The clusterName is displayed as hostname for KAC in the Host Management UI.
# Note: if the clusterName is detected by the KAC, this value will be overwritten
```

pflanagan-cs (approved):
lgtm

PR #268, opened 2024-02-22
Allow setting nodeSelector and affinity for falcon-kac

iandrewt:
This will allow falcon-kac to be used on multi-arch clusters, as long as there is a amd64 node available.

I've tested this is working on a fresh EKS cluster with Karpenter - it brought up a c7i.large instance alongside the m7g.large instances that were there already

### DIFF
--- a/helm-charts/falcon-kac/Chart.yaml
+++ b/helm-charts/falcon-kac/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.0.8
+version: 1.1.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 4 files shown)


redhatrises:
Thanks for opening this PR. Please update to use nodeAffinity to reference arch instead of nodeSelector and keep the nodeSelector the same.

iandrewt:
Sure. Would it be better to have the nodeSelector/nodeAffinity rules be hard-coded in `deployment_webhook.yaml` with values to add to them, or have them both in the values (as I have now)?

redhatrises:
> Sure. Would it be better to have the nodeSelector/nodeAffinity rules be hard-coded in `deployment_webhook.yaml` with values to add to them, or have them both in the values (as I have now)?

I would hard-code both of them, and then only the nodeaffinity should be allowed to have values added to them.

iandrewt:
I ended up putting the nodeAffinity default rule in the values as self-managing merging for that is more effort than it's worth. NodeSelector is reverted to being hard-coded in `deployment_webhook.yaml`

pflanagan-cs (approved):

PR #272, opened 2024-03-06
Increase KAC webhook timeout to 10 seconds

pflanagan-cs:
Increasing webhook timeout to 10 seconds to allow more time for ImageAssessment

### DIFF
--- a/helm-charts/falcon-kac/templates/deployment_webhook.yaml
+++ b/helm-charts/falcon-kac/templates/deployment_webhook.yaml
@@ -286,7 +286,7 @@ webhooks:
     - pods/ephemeralcontainers
     scope: '*'
   sideEffects: None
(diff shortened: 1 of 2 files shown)


helm-charts/falcon-kac/templates/deployment_webhook.yaml
```diff
@@ -359,4 +359,4 @@ webhooks:
     - jobs
     scope: '*'
   sideEffects: None
-  timeoutSeconds: 5
\ No newline at end of file
+  timeoutSeconds: 10
```
redhatrises, line 362: @pflanagan-cs can you add the End of File newline? Once that is added, it's good to merge.

pflanagan-cs: done

PR #273, opened 2024-03-12
Add support to falcon-sensor for existing secret in addition to configmap

nicolasbelanger:
Add support for an existing secret.  This is useful for one who wishes to use tools like external secret to provide the environment variables necessary (and prevent storing key/secret like clientID or provisioning token in git).

Skip injecting `falcon.cid` and `falcon.provisioning_token` in the configMap when `node.daemonset.existingSecret` is set.

### DIFF
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.26.1
+version: 1.26.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.26.1
+appVersion: 1.26.2
 
 keywords:
   - CrowdStrike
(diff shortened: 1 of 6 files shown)


ls-julien-deruere (approved):

opp-svega:
Any idea when this PR is going to get merged and deployed?

falcon-pioupiou:
don't think this PR will be merged as it's already supported since a while : 
https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/values.yaml#L81 

```
--set node.image.pullSecrets: <pullsecretname>
```

image

opp-svega:
@falcon-pioupiou Aren't `node.image.pullSecrets` different from the `envFrom` secrets for the container?

redhatrises:
Hello,

Thank you for the PR. Sorry for the delay. This PR introduces a security misconfig through the usage secretRef, so it won't be merged. Also cid and provisioning token are incredibly easy to discover wherever the sensor is installed regardless of it being k8s or windows, mac, etc. Closing this PR for this reason.

falcon-pioupiou:
@opp-svega sorry I've misread it and thought it was around pullsecrets only. 

@opp-svega @nicolasbelanger : how are you deploying today your helm ? CD pipeline, flux, argocd ?

to prevent storing key/secret like clientID or provisioning token in git you can do it by using sops, it will encrypt your helm values in your yaml using a regex like (here example with an azure Keyvault, but there is many options for sops) : 
```
sops --azure-kv $KEY_URL \
--encrypt --encrypted-regex '^(data|stringData|crowdstrikeConfig)$' \
--in-place apps/staging/values/falcon-sensor-values.yaml
```

All entries in data section will be encrypted (including the proxy config)

opp-svega:
We are using Argo right now, but nothing working currently for injecting the CID. I was hoping this PR was going to get merged before going down any other routes.

tskinn:
>This PR introduces a security misconfig through the usage secretRef

@redhatrises can you explain this more?
It seems very odd not to allow folks to use an existing secret. Why force people to use sops or any other solution besides the native secret implementation?

nicolasbelanger:
Sorry I'm late to the party...

@falcon-pioupiou We're deploying using ArgoCD, and storing all our secrets in Vault, pulled with [`external-secrets`](https://external-secrets.io/latest/).  We were hopeful to keep using this pattern on our infra with Cowdstrike, specially that you implemented the same pattern on some other charts of yours (ref.: https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-image-analyzer/templates/daemonset.yaml#L71-L76)

@redhatrises Can you share how `secretRef` is any less secure (or a worse configuration) than the current [`configMapRef`](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/daemonset.yaml#L175-L176)?

PR #274, opened 2024-03-14
cleanup: remove legacy code

redhatrises:
- cleanup(nodesensor): remove legacy initContainer code
- cleanup(sidecar): remove upgrade webhook as the service still functions with multiple replicas on upgrade


helm-charts/falcon-sensor/README.md
```diff
@@ -29,10 +29,8 @@ The Falcon Helm chart has been tested to deploy on the following Kubernetes dist
 
 | Helm chart Version      | Falcon Sensor Version             |
 |:------------------------|:----------------------------------|
-| `<= 1.6.x`              | `<= 6.34.x`                       |
-| `>= 1.7.x && <= 1.17.x` | `>= 6.35.x && < 6.49.x`           |
-| `>= 1.18.x`             | `>= 6.49.x`                       |
-| `>= 1.19.x`             | `>= 6.54.x`                       |
```
cs-alphonse, line 35: For my clarification; We removed the old versions, since those sensor versions are not supported?

helm-charts/falcon-sensor/templates/container_secret.yaml
```diff
@@ -12,7 +12,7 @@ data:
 type: kubernetes.io/dockerconfigjson
 {{- if or .Values.container.image.pullSecrets.namespaces .Values.container.image.pullSecrets.allNamespaces }}
 {{- $name := ( .Values.container.image.pullSecrets.name | default (printf "%s-pull-secret" (include "falcon-sensor.fullname" .))) }}
-{{- $myns := split "," .Values.container.image.pullSecrets.namespaces -}}
+{{- $myns := split "," .Values.container.image.pullSecrets.namespaces | default "" -}}
```
cs-alphonse, line 15: If _namespace_ is not provided and creating this resource in _default_ namespace, would it be in sync with other resources created by this helm chart instantiation?
(i,e) would other resources also be created in _default_ namespace?

helm-charts/falcon-sensor/README.md
redhatrises: Correct, yes.

helm-charts/falcon-sensor/templates/container_secret.yaml
redhatrises: Not sure what you are asking here, these settings are for deploying the secret to the non-deployment namespaces when the injector is used.

cs-alphonse: ACK. 
That clarified my query. I thought that this change was for the `injector` itself.

cs-alphonse (approved):

pflanagan-cs (approved):

PR #276, opened 2024-03-26
Forcing the param namespaces to be a default empty string

falcon-pioupiou:
if not forced to empty string, helm is taking the variable as an interface{}

without the default value : 
```
Error: template: falcon-sensor/templates/container_secret.yaml:15:30: executing "falcon-sensor/templates/container_secret.yaml" at <.Values.container.image.pullSecrets.namespaces>: wrong type for value; expected string; got interface {}
helm.go:84: [debug] template: falcon-sensor/templates/container_secret.yaml:15:30: executing "falcon-sensor/templates/container_secret.yaml" at <.Values.container.image.pullSecrets.namespaces>: wrong type for value; expected string; got interface {}
```

### DIFF
--- a/helm-charts/falcon-sensor/values.yaml
+++ b/helm-charts/falcon-sensor/values.yaml
@@ -188,7 +188,7 @@ container:
       # list. For example:
       #
       # namespaces: ns1,ns2,ns3
-      namespaces:
+      namespaces: ""
 
       # Attempt to create the Falcon sensor pull secret in all Namespaces
       # instead of using "container.image.pullSecrets.namespaces"
--- a/helm-charts/falcon-sensor/Chart.yaml
+++ b/helm-charts/falcon-sensor/Chart.yaml
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.26.1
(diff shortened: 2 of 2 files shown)


helm-charts/falcon-sensor/values.yaml
```diff
@@ -188,7 +188,7 @@ container:
       # list. For example:
       #
       # namespaces: ns1,ns2,ns3
-      namespaces:
+      namespaces: ""
```
redhatrises, line 191: This works. Could also set the following in `helm-charts/falcon-sensor/templates/container_secret.yaml` on line 15 that may or may not be better:

```
{{- $myns := split "," (.Values.container.image.pullSecrets.namespaces | default "") -}}
```

Thoughts?

falcon-pioupiou: I would go on the values.yaml file as other tools (like argocd) are reading the values.yaml to build a UI didn't checked if it's decoding well or not on argocd but I'm fine with the default.

I think it's a helm issue so could impact other tools reading the values.yaml without looking at the schema.

redhatrises: `{{- $myns := split "," (.Values.container.image.pullSecrets.namespaces | default "") -}}` should also solve your problem since the schema value and validation is `null` or `string` which means it can handle both and not error out. Schema validation will need to be fixed as well to be string only if pursuing updating the in-chart `values.yaml ` file.

PR #287, opened 2024-05-02
feat: configure nodeaffinity to properly support mixed-node architectures

redhatrises:


helm-charts/falcon-kac/templates/deployment_webhook.yaml
```diff
@@ -92,6 +92,23 @@ spec:
         {{- end }}
       {{- end }}
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: node-role.kubernetes.io/control-plane
```
pflanagan-cs, line 100: I think it's better not to add this expression for `node-role.kubernetes.io/control-plane`
Control plane nodes will have taint: `node-role.kubernetes.io/control-plane:NoSchedule` 
so control plane nodes will be excluded by taint without need for nodeAffinity
In general for our testing that taint is removed to allow KAC to be installed on single-node clusters
If this required nodeAffinity rule is added, removing the taint will no longer work

helm-charts/falcon-kac/templates/deployment_webhook.yaml
```diff
@@ -92,6 +92,23 @@ spec:
...
+                operator: DoesNotExist
+              - key: node-role.kubernetes.io/master
+                operator: DoesNotExist
+              - key: kubernetes.io/os
+                operator: In
+                values:
+                - linux
```
pflanagan-cs, line 107: This duplicates `nodeSelector`
https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-kac/templates/deployment_webhook.yaml#L209-L210

helm-charts/falcon-kac/templates/deployment_webhook.yaml
redhatrises: Good point. I actually dropped all the falcon-kac helm chart changes in favor of #268

helm-charts/falcon-kac/templates/deployment_webhook.yaml
redhatrises: Dropped. See previous comments.

pflanagan-cs (approved):

Issue #283, opened 2024-04-25
Integration Gateway does not take existing secret


mkeats-clumio:
Falcon integration gateway does not take an existing secret and expects the credentials to be passed in the values. This doesn't work well with certain gitops workflows.

https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-image-analyzer/values.yaml#L141 shows the correct way to do this. In general there is weak consistency between the charts in this repo with how they take configuration

PR #294, opened 2024-05-21
feat: add existing secret functionality to Falcon Integration Gateway

mkeats-clumio:
Fixes https://github.com/CrowdStrike/falcon-helm/issues/283

This adds a existingSecret value in the same style as the other charts in this repo.

It also adds the option to pass in arbitrary environment variables in order to manage the configurations that are embedded in the secret.


helm-charts/falcon-integration-gateway/templates/deployment.yaml
```diff
@@ -41,8 +41,14 @@ spec:
               mountPath: /fig/config/config.ini
               subPath: config.ini
           envFrom:
+          - configMapRef:
+              name: {{ include "falcon-integration-gateway.fullname" . }}-env
```
redhatrises, line 45: Drop this change.

helm-charts/falcon-integration-gateway/templates/configmap-env.yaml
```diff
@@ -0,0 +1,30 @@
...
+{{- end }}
+{{- if .Values.push.cloudtrail_lake.enabled }}
+  CLOUDTRAIL_LAKE_CHANNEL_ARN: {{ .Values.push.cloudtrail_lake.channel_arn | quote }}
+  CLOUDTRAIL_LAKE_REGION: {{ .Values.push.cloudtrail_lake.region | quote }}
+{{- end }}
+{{- if .Values.push.vmware_workspace_one.enabled }}
+  SYSLOG_HOST: {{ .Values.push.vmware_workspace_one.syslog_host | quote }}
+  SYSLOG_PORT: {{ .Values.push.vmware_workspace_one.syslog_port }}
+  WORKSPACEONE_TOKEN: {{ .Values.push.vmware_workspace_one.token | quote }}
+{{- end }}
```
redhatrises, line 30: This should stay where it originally was in a secret since it still contains sensitive data. So please drop this configmap addition as well as the configmap file rename.

redhatrises (changes requested):

helm-charts/falcon-integration-gateway/templates/configmap-env.yaml
mkeats-clumio: The intent is to move the non-sensitive data out of the secret and into a config map where it belongs. If I set up an existing secret I don't want to be managing `SYSLOG_HOST` there for example. It should only contain sensitive data so I can keep the config managed in helm.

Another option is to use an `env:` key values

It looks like missed `GOOGLE_SECURITY_KEY` and WORKSPACEONE_TOKEN however. Can you clarify if anything else is sensitive?

redhatrises: You'd want to update the config.ini and not create a new configmap. Technically, secrets are not secrets in k8s, and the use of `secretRef` should be avoided due to secrets as env var exposure, credential dumping, and exporting plaintext via logs. There are other sensitive values like `WORKSPACE_ID` and `PRIMARY_KEY`

mkeats-clumio: See changes

mkeats-clumio: The existing chart uses secretRef I just made the name here configurable.

In order to reduce scope of changes I moved to use `env` instead of attempting to figure out what was correctly a secret or not.  This allows me to set AWS_REGION in the vars as well as the other missing env vars.

redhatrises:
@mkeats-clumio see if #297 fits your needs. Turns out there was a good amount of updating that needed to happen for the config. I have also added your existing secret code there.

mkeats-clumio:
@redhatrises that looks perfect. I'll test it out shortly. I'll close this PR after I confirm

mkeats-clumio:
#297 Looks good, closing this one

PR #297, opened 2024-06-14
chore: update FIG chart to the latest released version 3.2.0

redhatrises:


helm-charts/falcon-integration-gateway/templates/configmap.yaml
```diff
@@ -20,38 +22,210 @@ data:
...
+    [secrets_manager]
+    # Uncomment to provide aws region for Secrets Manager. Alternatively, use SECRETS_MANAGER_REGION env variable.
+    #region = us-west-2
+    {{- if .Values.credentials_store.secrets_manager.region }}
+    region = {{ .Values.credentials_store.secrets_manager.region }}
+    {{- end }}
+
+    # Uncomment to provide Secrets Manager secret name. Alternatively, use SECRETS_MANAGER_SECRET_NAME env variable.
+    #secrets_manager_secret_name = falcon/fig/credentials
+
```
carlosmmatos, line 113:
```suggestion
    {{- if .Values.credentials_store.secrets_manager.secret_name }}
    secrets_manager_secret_name = {{ .Values.credentials_store.secrets_manager.secret_name }}
    {{- end }}
```

carlosmmatos (changes requested):
Outside of the requestd change, I think it looks good

helm-charts/falcon-integration-gateway/templates/configmap.yaml
redhatrises: Done

carlosmmatos (approved):

mkeats-clumio:
@redhatrises there wasn't a 0.4.0 version published in the helm repo for this change. The latest is still 0.3.0
The github action failed due to some unrelated tag already existing https://github.com/CrowdStrike/falcon-helm/actions/runs/9551134785

mkeats-clumio:
@redhatrises looks like it published after I posted that.

Testing the new version I found a problem with the way you templated out the config

```
    #confirm_instance = true
    {{- if .Values.push.aws_security_hub.confirm_instance }}
    confirm_instance = {{ .Values.push.aws_security_hub.confirm_instance }}
    {{- end }}

```
`{{- if .Values.push.aws_security_hub.confirm_instance }}` is only evaluated true if confirm_instance is set to true. So it makes it impossible to ever override the default true value which I found out when trying to set it to false. We want to check if the value is defined, not if it is true. 

It would be simpler just to remove the conditional and let confirm_instance pick up default value of true if not overridden. This would apply for the other vars too.

PR #327, opened 2024-10-18
Add support for running KAC on hostNetwork

TomRyan-321:
This PR introduces support for running KAC on hostNetwork and setting of dnsPolicy which is needed when some scenarios such as when custom CNI's are in use and control plane nodes are not be able to initiate network connections to pods.

Additionally this change adds support for configuring the watcher port and fixes missing webhookPort from being passed to the container client.

PR #331, opened 2024-10-30
v1.1.11 helm updates for hostNetwork

cs-pvyas:
v1.1.11 helm updates for hostNetwork


cs-pvyas:
@gpontejos @mhyson-cs  - Do we need to introduce the `dnsPolicy` value as well so that customer can specify that ?
Similar to https://github.com/CrowdStrike/falcon-helm/pull/327/files ?????

gpontejos:
> @gpontejos @mhyson-cs - Do we need to introduce the `dnsPolicy` value as well so that customer can specify that ? Similar to https://github.com/CrowdStrike/falcon-helm/pull/327/files ?????

@cs-pvyas Yes, that'll need to be added if the customer is using the hostNetwork:  https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy

gpontejos (approved):

PR #335, opened 2024-11-15
relax CID validation pattern

krutsko:
The CID with used doesn't match the pattern from the helm chart, so we need to apply dirty hack to workaround this.

### DIFF
--- a/helm-charts/falcon-sensor/values.schema.json
+++ b/helm-charts/falcon-sensor/values.schema.json
@@ -10,7 +10,7 @@
             "properties": {
                 "cid": {
                     "type": "string",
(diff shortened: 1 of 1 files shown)


ChristianCiach:
We have a similar workflow in place. But you should be able to call `helm template --skip-schema-validation` instead, which I believe to be the better solution.

krutsko:
Yes, probably, but our current tooling does not allow to call `helm template --skip-schema-validation`. Also, I don't want to skip validation altogether, it's useful. 
But, instead, I think it makes sense to relax the cumbersome validation pattern. It means, I don't forget to set some value here (if it doesn't correct, it's my fault in the end). And at the same time, all other fields will pass validation according to the scheme.

PR #340, opened 2024-12-04
sensor | clusterrole | add resources and verbs

JonBruchim:
partial output of 

`kubectl get clusterrole falcon-falcon-sensor-access-role  -n falcon-system -o yaml`

```

    app.kubernetes.io/component: kernel_sensor
    app.kubernetes.io/instance: falcon
    
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  - pods
  - services
  - nodes
  - daemonsets
  - replicasets
  - deployments
  - jobs
  - ingresses
  - cronjobs
  - persistentvolumes
  verbs:
  - get
  - watch
  - list
 ```

### DIFF
--- a/helm-charts/falcon-sensor/templates/clusterrole.yaml
+++ b/helm-charts/falcon-sensor/templates/clusterrole.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.container.enabled }}
+{{- if or .Values.container.enabled .Values.node.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -8,7 +8,11 @@ metadata:
     app.kubernetes.io/name: {{ include "falcon-sensor.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{ if .Values.container.enabled }}
     app.kubernetes.io/component: "container_sensor"
+    {{ else if .Values.node.enabled }}
+    app.kubernetes.io/component: "kernel_sensor"
+    {{ end }}
     crowdstrike.com/provider: crowdstrike
     helm.sh/chart: {{ include "falcon-sensor.chart" . }}
 rules:
@@ -16,6 +20,22 @@ rules:
   - ""
   resources:
   - secrets
+  {{- if and .Values.node.enabled }}
+  - pods
+  - services
+  - nodes
+  - daemonsets
+  - replicasets
+  - deployments
+  - jobs
+  - ingresses
+  - cronjobs
+  - persistentvolumes
+  {{- end }}
   verbs:
   - get
+  {{- if .Values.node.enabled }}
+  - watch
+  - list
+  {{- end }}
 {{- end }}
--- a/helm-charts/falcon-sensor/templates/clusterrolebinding.yaml
+++ b/helm-charts/falcon-sensor/templates/clusterrolebinding.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.container.enabled }}
+{{- if or .Values.container.enabled .Values.node.enabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: helm-charts/falcon-sensor/README.md [hunk 1]
```diff
@@ -127,7 +127,8 @@ The following tables lists the more common configurable parameters of the chart
...
 | `node.image.registryConfigJSON`   | base64 encoded docker config json for the pull secret                  | None       (Conflicts with node.image.pullSecrets)                      |
 | `node.daemonset.resources`        | Configure Node sensor resource requests and limits (eBPF mode only)    | None       (Minimum setting of 250m CPU and 500Mi memory allowed). Default for GKE Autopilot is 750m CPU and 1.5Gi memory.<br><br><div class="warning">:warning: **Warning**:<br>If you configure resources, you must configure the CPU and Memory Resource requests and limits correctly for your node instances for the node sensor to run properly!</div> |
-| `falcon.cid`                      | CrowdStrike Customer ID (CID)                                          | None       (Required)                                                   |
+| `node.cdpRolesEnabled`            | Enabled the Cloud Data Protection module                               | `true`                                                                  |
```
cs-alphonse, line 130:
```suggestion
| `node.cdpRolesEnabled`            | Enable cluster roles for Cloud Data Protection module                               | `true`                                                                  |
```
Since this flag only allows the role to be created. Additional config changes(Falcon UI) are required to enable the protection module.

helm-charts/falcon-sensor/templates/clusterrole.yaml
```diff
@@ -16,6 +16,22 @@ rules:
   - ""
   resources:
   - secrets
+  {{- if and .Values.node.enabled .Values.node.cdpRolesEnabled }}
```
cs-alphonse, line 19: * I am not able to comment on content that is not part of the diff
* https://github.com/CrowdStrike/falcon-helm/pull/340/files#diff-8b1b64050a16b4744833a69c1dd56929ef1e6ef30cbc080ce0d4018c94a46f28R11
* https://github.com/CrowdStrike/falcon-helm/pull/340/files#diff-6ce3e7fab19d7aa64a73efef8b41d75116dc829eb636e3e22cfce3984e99d308L11
* `app.kubernetes.io/component` should be updated to `all_sensor`.
   * Not in the scope of this PR
      * We should replace `kernel_sensor` as `node_sensor`. We never updated the component after introducing support for `bpf`. Adding @gpontejos

helm-charts/falcon-sensor/templates/clusterrole.yaml
```diff
@@ -16,6 +16,22 @@ rules:
   - ""
   resources:
   - secrets
```
cs-alphonse, line 18: Is `secrets` required by `cdp`?

JonBruchim: Yes

helm-charts/falcon-sensor/templates/clusterrole.yaml
cs-alphonse: Checked other resources that we create; `all_sensor` is applicable, only if we use the same resource name for multiple sensor types. In this case we create resource using the variable `"falcon-sensor.fullname"` and that is unique per deployment. So we should add
```
{{ if .Values.container.enabled }}
app.kubernetes.io/component: "container_sensor"
{{ else if (and .Values.node.enabled .Values.node.cdpRolesEnabled)}}
app.kubernetes.io/component: "kernel_sensor"
{{ end }}
```

helm-charts/falcon-sensor/values.yaml
```diff
@@ -104,6 +104,9 @@ node:
     postDelete:
       enabled: true
 
+  # When enabled, Cloud data protection module is enabled
+  cdpRolesEnabled: true
```
redhatrises, line 108: If data protection is not a separate sku, why are we adding this boolean? Permissions should be applied by default.

redhatrises:
@gpontejos good to merge? We can kill the release trigger or do you want a new release?

gpontejos (approved):

PR #349, opened 2025-01-16
feat: only run release if label is applied to PR

redhatrises:


.github/workflows/helm-release.yml
```diff
@@ -1,14 +1,16 @@
...
+  pull_request:
+    types: [closed]
     branches:
       - main
     paths:
       - 'helm-charts/**'
 
 jobs:
   release:
+    if: github.event.pull_request.merged == true && contains(github.event.pull_request.labels.*.name, 'release')
```
mhyson-cs, line 13: Is `merged` a field of the pull-request object? I'm not seeing it in the [reference](https://docs.github.com/en/rest/pulls). There's `merged_at` and `merge_commit_sha`, but nothing that just says `merged`.

gpontejos: It's the github actions docs [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges).

mhyson-cs: Weird. Inconsistency in the docs I guess.

gpontejos (approved):

PR #350, opened 2025-01-16
fix: only run version bump check if release label is present in PR

gpontejos:


redhatrises:
@gpontejos these are actually meant to run outside of a release.

redhatrises (approved):

PR #356, opened 2025-01-27
feat: Make securityContext configurable

papanito:
Also extend with seccompProfile and capabilities as defaults


papanito:
@mcrowson could you please have a look at this?

mcrowson:
These generally look good, thanks for the contribution! We'll run some tests with these changes locally and look to incorporate them in a future release.

papanito:
@mcrowson  something missing? would really appreciate to include this change so we can start using the official chart rather than maintaining our own

mcrowson:
This was released as part of 1.6.0 https://falcon.crowdstrike.com/support/release-notes/release-notes-self-hosted-registry-assessment-shra-version-1-6-0

PR #367, opened 2025-02-20
feat: Allow disable ValidatingWebhookConfiguration for kac

papanito:
If you don't need the validationWebhook cause you already have another application in place you can disable it completely


papanito:
Apparently the config is still needed otherwise pod does not start

```
E0220 12:17:15.496113 112] "Failed to read webhook metadata" "error"="KAC webhook configuration not found"
```

PR #378, opened 2025-03-19
Add alternateMountPath option to Falcon Helm

cs-bqin02:


helm-charts/falcon-sensor/questions.yaml
```diff
@@ -185,6 +185,14 @@ questions:
     label: Disable Pod injection
     group: "Falcon Container settings"
 
+  - variable: container.alternateMountPath
+    description: "Enable mount path at /falcon for NVCF"
```
cs-alphonse, line 189: Please use the same help string as defined in `installer` to keep it consistent
"Enable volume mounts at /falcon instead of /tmp for NVCF environment"

cs-alphonse (approved):

PR #387, opened 2025-04-14
Add ASPM Relay helm chart

olagziel:


redhatrises:
@olagziel please fix all the files missing there end of file new lines

helm-charts/aspm-relay/templates/role.yaml
```diff
@@ -28,4 +28,5 @@ rules:
   - apiGroups: [""]
     resources: ["pods/exec"]
     verbs: ["get", "create"]
-  {{- end -}}
\ No newline at end of file
+  {{- end -}}
+
```
redhatrises, line 32: @olagziel seems like this file still has it.

olagziel: Do you want me to remove or add newlines at the end of these files?

olagziel: fixed

PR #393, opened 2025-04-24
Add AllowListSynchronizer for Autopilot workloads

gpontejos:


helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -142,3 +142,15 @@ args:
       test -f "/opt/CrowdStrike/configure-cluster-id" && /opt/CrowdStrike/configure-cluster-id || echo "/opt/CrowdStrike/configure-cluster-id not found. Skipping."
 {{- end -}}
 {{- end -}}
+
+{{/*
+Check if target cluster supports GKE Autopilot WorkloadAllowlists.
+Documentation: https://cloud.google.com/kubernetes-engine/docs/reference/crds/allowlistsynchronizer
+*/}}
+{{- define "gke-autopilot-workloadallowlists-enabled" -}}
+{{- if and .Values.node.gke.autopilot (.Capabilities.APIVersions.Has "auto.gke.io/v1/AllowlistSynchronizer") (.Capabilities.APIVersions.Has "auto.gke.io/v1/WorkloadAllowlist") (semverCompare ">=v1.32.0-gke.1000000" .Capabilities.KubeVersion.Version) -}}
```
cs-alphonse, line 151: There was a bug fixed by AP team in `1.32.2-gke.16520000` w.r.t `matching-allowlist`.
Let's mention that has the min required version to avoid Cx running into that issue

helm-charts/falcon-sensor/templates/gke_autopilot_allowlist_synchronizer.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if eq (include "gke-autopilot-workloadallowlists-enabled" .) "true" }}
```
cs-alphonse, line 1: This should be enabled only if `node.gke.autopilot` is set to `true`.
Also, could we throw an error message it below condition is true
* node.gke.autopilot is true
* gke-autopilot-workloadallowlists-enabled is false

Since all future allow-listing is going to be via this mechanism, if we do not throw an error(due to the above condition being false), the creation of DaemonSet resource would fail with an error(that might be confusing to Cx)

helm-charts/falcon-sensor/templates/gke_autopilot_allowlist_synchronizer.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: crowdstrike-synchronizer
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ include "falcon-sensor.chart" . }}
+  annotations:
+    "helm.sh/hook-weight": "-999"
+spec:
+  allowlistPaths:
+  - CrowdStrike/falcon-sensor/*
```
cs-alphonse, line 13: I was using `*` for testing.
Let's go ahead with the below list, since we know the versions that has the required allow-list and the corresponding image digests for private registries
```
- CrowdStrike/falcon-sensor/cleanup/crowdstrike-falconsensor-cleanup-allowlist-v1.0.1.yaml
- CrowdStrike/falcon-sensor/deploy/crowdstrike-falconsensor-deploy-allowlist-v1.0.0.yaml
- CrowdStrike/falcon-sensor/deploy/crowdstrike-falconsensor-deploy-allowlist-v1.0.2.yaml
- CrowdStrike/falcon-sensor/falconctl/crowdstrike-falconsensor-falconctl-allowlist-v1.0.1.yaml
```

Note: The reason we have two versions for deploy; This is to support sensor version with both w/ and w/o `cluster-id`.

cs-alphonse (changes requested):

rindeastwood:
What happened to this wonderful idea of packaging the allowlist into the chart? :-)

PR #394, opened 2025-04-24
Add nodeAffinity to sidecar sensor

gpontejos:


helm-charts/falcon-sensor/values.yaml
```diff
@@ -156,6 +156,17 @@ container:
   # Configure the Injector Port
   injectorPort: 4433
 
+  # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity
+  # Allow setting additional node selections e.g. processor type
+  # amd64 is the only supported architecture at this time.
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+      - matchExpressions:
```
cs-alphonse, line 165: Any reason to remove the match for `linux`? 
IIUC, we had added this, since some of the K8s cluster were running multiple Operating Systems.

gpontejos: @cs-alphonse Fixed.

PR #401, opened 2025-05-09
Ensure configurations work with WorkloadAllowlist

gpontejos:
- Add labels for troubleshooting the primary and cleanup daemonsets for AutoPilot.
- Set configMap name to be a static value when Autopilot is enabled

### DIFF
--- a/helm-charts/falcon-sensor/templates/_helpers.tpl
+++ b/helm-charts/falcon-sensor/templates/_helpers.tpl
@@ -134,11 +134,37 @@ args:
   - '-c'
   - >-
       echo "Running /opt/CrowdStrike/falcon-daemonset-init -i";
-{{- if .Values.node.gke.autopilot }}
-      /opt/CrowdStrike/falcon-daemonset-init -i
-{{- else -}}
       /opt/CrowdStrike/falcon-daemonset-init -i;
       echo "Running /opt/CrowdStrike/configure-cluster-id";
       test -f "/opt/CrowdStrike/configure-cluster-id" && /opt/CrowdStrike/configure-cluster-id || echo "/opt/CrowdStrike/configure-cluster-id not found. Skipping."
+{{- end -}}
+
+{{/*
+Create the name of the config map if GKE Autopilot is used.
+WorkloadAllowlists require an exact match for naming.
+*/}}
+{{- define "falcon-sensor.configMapName" -}}
+{{- if and .Values.node.gke.autopilot -}}
+{{- printf "falcon-node-sensor-config" -}}
+{{- else -}}
+{{- printf "%s-config" (include "falcon-sensor.fullname" .) -}}
+{{- end -}}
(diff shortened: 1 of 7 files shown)


helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -134,11 +134,37 @@ args:
...
       echo "Running /opt/CrowdStrike/configure-cluster-id";
       test -f "/opt/CrowdStrike/configure-cluster-id" && /opt/CrowdStrike/configure-cluster-id || echo "/opt/CrowdStrike/configure-cluster-id not found. Skipping."
 {{- end -}}
+
+{{/*
+Create the name of the config map if GKE Autopilot is used.
+WorkloadAllowlists require an exact match for naming.
+*/}}
+{{- define "falcon-sensor.configMapName" -}}
+{{- if and .Values.node.gke.autopilot .Values.node.enabled -}}
```
cs-alphonse, line 147: nit: Could we have autopilot enabled without node set to true?
(i,e) do we need to check only for autopilot above?

helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
```diff
@@ -146,7 +146,7 @@ spec:
         command: ["injector"]
         envFrom:
         - configMapRef:
-            name: {{ include "falcon-sensor.fullname" . }}-config
+            name: {{ include "falcon-sensor.configMapName" . }}
```
cs-alphonse, line 149: Please ensure to run Lumos regression also

helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -134,11 +134,37 @@ args:
   - '-c'
   - >-
       echo "Running /opt/CrowdStrike/falcon-daemonset-init -i";
-{{- if .Values.node.gke.autopilot }}
```
cs-alphonse, line 137: Are we including cluster-id changes also as part of this PR?

cs-alphonse (approved):

helm-charts/falcon-sensor/templates/_helpers.tpl
gpontejos: @cs-alphonse Yes, I planned on it so that the allowlistworkload would match.

helm-charts/falcon-sensor/templates/_helpers.tpl
gpontejos: @cs-alphonse I didn't want to affect any existing functionality with the container sensor.  I can remove the requirement, if needed.

gpontejos: @cs-alphonse Updated.  I forgot that container sensor field was gcp and not gke in the values.

PR #404, opened 2025-05-14
Add scaffolding basic tests for falcon-sensor and falcon-kac

gpontejos:


mr-jungchoi (approved):
LGTM, we should also add a GHA workflow to run the tests as a PR check.

PR #409, opened 2025-05-22
add checksums to the deployments of falcon-kac and falcon-sensor.

nobletrout:
Creates an annotation on deployment yaml for kac and sensor to force a restart of pods on configmap changes  (you change your CID, add new tags, etc).

Otherwise you need to run around restarting a bunch of a pods which gets old fast.

this impacted us using argocd:
https://github.com/argoproj/argo-cd/issues/13642

In which the developers are big meanie pants and tell me to go figure out my own problems instead providing a fix in their CICD system.


gpontejos (approved):
Apologies for the delay.  Looks good.  It'll be included in the next release.

PR #416, opened 2025-06-24
Add ability to run the cleanup Daemonset only

gpontejos:
Setting `node.cleanupOnly` to `true` will run the cleanup daemonset when `node.enabled`.  The Daemonset will run indefinitely and only be removed during a helm uninstall to ensure cleanup runs on all nodes.

### DIFF
--- a/helm-charts/falcon-sensor/values.schema.json
+++ b/helm-charts/falcon-sensor/values.schema.json
@@ -5,13 +5,6 @@
         "falcon": {
             "type": "object",
             "properties": {
-                "cid": {
-                    "type": "string",
-                    "pattern": "^[0-9a-fA-F]{32}-[0-9a-fA-F]{2}$",
-                    "example": [
-                        "1234567890ABCDEF1234567890ABCDEF-12"
-                    ]
-                },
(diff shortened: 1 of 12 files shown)


helm-charts/falcon-sensor/values.schema.json
```diff
@@ -521,41 +514,64 @@
...
+                            "falconSecret": {
+                                "properties": {
+                                    "enabled": { "enum": [true] },
+                                    "secretName": {
+                                        "type": "string",
+                                        "minLength": 1,
+                                        "pattern": "^[0-9a-fA-F]{32}-[0-9a-fA-F]{2}$",
+                                        "example": [
+                                            "1234567890ABCDEF1234567890ABCDEF-12"
+                                        ]
```
mr-jungchoi, line 556: lines 553-556 are for `cid`, not `secretName`. It also looks like the indentation is not aligned properly on lines 540-543.

mr-jungchoi (changes requested):
I think `node_secret.yaml` and `node_secret_cleanup.yaml` can also be excluded for clean up only.

PR #421, opened 2025-07-16
fix: adjust the whitespace chomp to ensure the following annotation d…

opp-svega:
Changing the whitespace chomp to ensure the annotations do not end up on the same line. This was causing issues with helm rendering the templates.

Error Message:
```
Use --debug flag to render out invalid YAML
Error: YAML parse error on falcon-sensor/templates/configmap.yaml: error converting YAML to JSON: yaml: line 10: mapping values are not allowed in this context
```


tcirakyopploans:
Yes, please fix this

gpontejos (approved):

PR #423, opened 2025-07-18
fix: Ensure correct nodeAffinity key is used based on k8s version

gpontejos:


helm-charts/falcon-sensor/templates/container_deployment_webhook.yaml
```diff
@@ -92,7 +92,7 @@ spec:
           - weight: 1
             preference:
               matchExpressions:
-              - key: node-role.kubernetes.io/master
+              - key: {{ if semverCompare "<1.24-0" .Capabilities.KubeVersion.Version }}node-role.kubernetes.io/master{{ else }}node-role.kubernetes.io/control-plane{{ end }}
```
pflanno, line 95: Isn't standard control plane node taint preventing pod from being scheduled on master node.

PR #425, opened 2025-07-18
fix: Remove unnecessary nodeAffinity expression from Lumos deployment

gpontejos:


helm-charts/falcon-sensor/values.yaml
```diff
@@ -169,18 +169,20 @@ container:
   # https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity
   # Allow setting additional node selections e.g. processor type
   # amd64 is the only supported architecture at this time.
```
pflanno, line 171: comment is no longer true, arm64 also supported
Will arm64 be added in a different PR?

gpontejos: @pflanno Good catch.  arm64 is supported so I removed the comment.

pflanno (approved):

PR #427, opened 2025-07-29
falcon-sensor: Add support for dnsConfig parameter

nitinverma9:


gpontejos:
Hi! I'll get this in next release. Can you remove any KPA updates and the version/app bumps from the sensor please? KPA has been deprecated and we're removing them from main here shortly.

PR #431, opened 2025-08-07
v1.1.16 IAR-KAC service port + readme

cs-pvyas:
v1.1.16 IAR-KAC service port + readme

### DIFF
--- a/helm-charts/falcon-image-analyzer/templates/daemonset.yaml
+++ b/helm-charts/falcon-image-analyzer/templates/daemonset.yaml
@@ -1,4 +1,36 @@
 {{- if .Values.daemonset.enabled }}
+{{- $name := (printf "%s" (include "falcon-image-analyzer.iarAgentService" .)) -}}
+{{- $fullName := (printf "%s.%s.svc" $name .Release.Namespace) -}}
+{{- if .Values.iarAgentService.domainName }}
+{{- $fullName = (printf "%s.%s.svc.%s" $name .Release.Namespace .Values.domainName) -}}
+{{- end }}
+{{- $myns := list }}
+{{- $certValid := (.Values.iarAgentService.certExpiration | int) -}}
+{{- $altNames := list ( printf "%s" $fullName ) ( printf "%s.cluster.local" $fullName ) ( printf "%s.%s" $fullName .Release.Namespace ) ( printf "%s" $fullName ) -}}
+{{- $ca := genCA ( printf "%s ca" .Release.Namespace ) $certValid -}}
+{{- $cert := genSignedCert $fullName nil $altNames $certValid $ca -}}
+{{- if not .Values.iarAgentService.autoCertificateUpdate }}
+{{- $tlscrt := (lookup "v1" "Secret" .Release.Namespace (printf "%s-tls" (include "falcon-image-analyzer.name" .))).data -}}
+{{- if kindIs "map" $tlscrt }}
+{{- $cert = dict "Cert" (index $tlscrt "tls.crt" | b64dec ) "Key" (index $tlscrt "tls.key" | b64dec ) -}}
+{{- end }}
+{{- end }}
+{{- $tlsCert := $cert.Cert | b64enc }}
+{{- $tlsKey := $cert.Key | b64enc }}
+{{- $caCert := $ca.Cert | b64enc -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "falcon-image-analyzer.name" . }}-tls
+  namespace: {{ .Release.Namespace }}
+  labels:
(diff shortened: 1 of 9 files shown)


helm-charts/falcon-image-analyzer/templates/daemonset.yaml
```diff
@@ -1,6 +1,38 @@
 {{- if .Values.daemonset.enabled }}
 {{- $imagePullSecret := include "falcon-image-analyzer.imagePullSecret" . }}
 {{- $registryConfigJson := include "falcon-image-analyzer.registryConfigJson" . }}
+{{- $name := (printf "%s" (include "falcon-image-analyzer.iarAgentService" .)) -}}
+{{- $fullName := (printf "%s.%s.svc" $name .Release.Namespace) -}}
```
mr-jungchoi: We will need to update all references to `.Release.Namespace` to something like:

```suggestion
{{- $iarNamespace := include "falcon-image-analyzer.namespace" . }}
{{- $fullName := (printf "%s.%s.svc" $name $iarNamespace) -}}
```

And then reuse that variable in this template, rather than using `.Release.Namespace` directly.

helm-charts/falcon-image-analyzer/templates/daemonset.yaml
```diff
@@ -1,6 +1,38 @@
...
+{{- $cert = dict "Cert" (index $tlscrt "tls.crt" | b64dec ) "Key" (index $tlscrt "tls.key" | b64dec ) -}}
+{{- end }}
+{{- end }}
+{{- $tlsCert := $cert.Cert | b64enc }}
+{{- $tlsKey := $cert.Key | b64enc }}
+{{- $caCert := $ca.Cert | b64enc -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "falcon-image-analyzer.name" . }}-tls
```
mr-jungchoi: suggestion: This seems to be duplicated in both the `daemonset.yaml` and `deployment.yaml` templates. We should extract the TLS secret into it's own template file so we aren't managing it in 2 different places.

mr-jungchoi (changes requested):
2 critical change requests for:
- ensuring namespace metadata is compatible with unified Helm chart
- ensuring the new TLS cert does not break Helm installation for ArgoCD users

helm-charts/falcon-image-analyzer/templates/daemonset.yaml
```diff
@@ -1,6 +1,38 @@
...
+{{- if .Values.iarAgentService.domainName }}
+{{- $fullName = (printf "%s.%s.svc.%s" $name .Release.Namespace .Values.domainName) -}}
+{{- end }}
+{{- $myns := list }}
+{{- $certValid := (.Values.iarAgentService.certExpiration | int) -}}
+{{- $altNames := list ( printf "%s" $fullName ) ( printf "%s.cluster.local" $fullName ) ( printf "%s.%s" $fullName .Release.Namespace ) ( printf "%s" $fullName ) -}}
+{{- $ca := genCA ( printf "%s ca" .Release.Namespace ) $certValid -}}
+{{- $cert := genSignedCert $fullName nil $altNames $certValid $ca -}}
+{{- if not .Values.iarAgentService.autoCertificateUpdate }}
+{{- $tlscrt := (lookup "v1" "Secret" .Release.Namespace (printf "%s-tls" (include "falcon-image-analyzer.name" .))).data -}}
```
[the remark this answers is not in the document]
cs-pvyas: This  can be fixed later. for now this is imp for this release

helm-charts/falcon-image-analyzer/templates/daemonset.yaml
cs-pvyas: Lets do the clean up post this release . As this was tested as is

helm-charts/falcon-image-analyzer/templates/daemonset.yaml
cs-pvyas: Lets do the clean up post this release . As this was tested as is

helm-charts/falcon-image-analyzer/values.schema.json
```diff
@@ -133,7 +133,7 @@
         "crowdstrikeConfig": {
             "type": "object",
             "required": [
-                "clusterName"
+                "cid"
```
mr-jungchoi: `cid` should not be required here in the schema. This will break the unified Helm chart's usage of `global.falcon.cid`. Instead of adding `cid` as required here. I've added validation helpers in the other Helm charts to account for either an inherited global CID or the component specific CID.

ref: https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-kac/templates/_helpers.tpl#L189-L204

PR #432, opened 2025-08-08
feat(falcon-sensor): add support for falcon.cloud

mr-jungchoi:


helm-charts/falcon-sensor/values.schema.json
```diff
@@ -5,6 +5,13 @@
         "falcon": {
             "type": "object",
             "properties": {
+                "cloud": {
+                    "type": [
+                        "null",
+                        "string"
+                    ],
+                    "enum": [null, "", "us-1", "us-2", "eu-1", "us-gov-1", "us-gov-2"]
```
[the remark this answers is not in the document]
mr-jungchoi: [Null and empty values for `falcon` options get filtered out](https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-sensor/templates/configmap.yaml#L31) before including them in the configmap. However, we've had some customer issues where empty values are being set even if they try to assign null values in the manifest, which I believe is dependent on their method of deploying the helm charts. For this reason I've included `""` even though it won't actually apply this value to the configmap and subsequently be applied to the Falcon sensor installation options.

For additional context `""` is also an allowed value for `trace`, but that also get filtered out.

helm-charts/falcon-sensor/README.md
```diff
@@ -114,22 +114,23 @@ For more details please see the [falcon-helm](https://github.com/CrowdStrike/fal
...
+| `falcon.cid`                    | CrowdStrike Customer ID (CID)                                                                                                                                                                                                                                                                         | None       (Required if falconSecret.enabled is false)                                                                     |
+| `falcon.cloud`                  | CrowdStrike cloud region (`us-1`, `us-2`, `eu-1`, `us-gov-1`, `us-gov-2`)<br><br>**NOTE:** `falcon.cloud` is only supported by Falcon Sensor version 7.28+                                                                                                                                            | None                                                                                                                       |
```
cs-alphonse, line 131: nit:
Replace the note as `This option is supported by Falcon sensor version 7.28 and above`

helm-charts/falcon-sensor/README.md
```diff
@@ -114,22 +114,23 @@ For more details please see the [falcon-helm](https://github.com/CrowdStrike/fal
...
+| `falcon.cid`                    | CrowdStrike Customer ID (CID)                                                                                                                                                                                                                                                                         | None       (Required if falconSecret.enabled is false)                                                                     |
+| `falcon.cloud`                  | CrowdStrike cloud region (`us-1`, `us-2`, `eu-1`, `us-gov-1`, `us-gov-2`)<br><br>**NOTE:** This option is supported by Falcon sensor version 7.28 and above                                                                                                                                           | None                                                                                                                       |
```
cs-alphonse, line 131: Posting for reference; This option would be support container sensor and KAC in a future release.

cs-alphonse (approved):

PR #446, opened 2025-09-18
Support KAC communication with Falcon Image Analyzer

pflanagan-cs:
KAC version 7.31 can communicate with Falcon Image Analyzer (IAR) to request Image Analalysis
KAC needs to know the namespace of Falcon Image Analyzer (IAR) in order to discover if IAR is installed.
It assumes namespace of "falcon-image-analyzer" but there needs to be a way to override that namespace if Falcon Image Analyzer (IAR) is installed into a different namespace
KAC also needs permission to read CA cert from IAR tls secret


helm-charts/falcon-kac/Chart.yaml
```diff
@@ -15,12 +15,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.5.1
+version: 1.5.2
```
gpontejos: We'll do the version bump in a separate PR. 
```suggestion
version: 1.5.1
```

helm-charts/falcon-kac/Chart.yaml
```diff
@@ -15,12 +15,12 @@ type: application
...
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.5.1
+appVersion: 1.5.2
```
gpontejos: We'll do the version bump in a separate PR.
```suggestion
appVersion: 1.5.1
```

PR #463, opened 2025-10-30
fix(falcon-sensor): Prevent daemonset install when container image is used

gpontejos:


redhatrises (approved):

helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -133,6 +133,11 @@ Create init script for daemonset
 args:
   - '-c'
   - >-
+      set -e;
+      if [ ! -f /opt/CrowdStrike/falcon-daemonset-init ]; then
+        echo "The container image is not a falcon sensor(DaemonSet) container image";
```
cs-alphonse:
```suggestion
        echo "Error: This is not a falcon node sensor(DaemonSet) image";
```

PR #469, opened 2025-11-10
add FALCON_CLUSTER_NAME env var for unmanages k8s clusters

JonBruchim:


helm-charts/falcon-sensor/templates/configmap.yaml
```diff
@@ -36,6 +36,7 @@ data:
 
   {{- if and .Values.node.enabled .Values.node.backend }}
   FALCONCTL_OPT_BACKEND: "{{ .Values.node.backend }}"
+  FALCON_CLUSTER_NAME: "{{ .Values.node.clusterName }}"
   {{- end }}
```
mr-jungchoi:
```suggestion
{{- if .Values.node.enabled }}
{{- if .Values.node.backend }}
FALCONCTL_OPT_BACKEND: "{{ .Values.node.backend }}"
{{- end }}
{{- if .Values.node.clusterName }}
FALCON_CLUSTER_NAME: "{{ .Values.node.clusterName }}"
{{- end }}
{{- end }}
```

Issue #462, opened 2025-10-30
Typo: Should be image.repository


jmckenzie-cs:
Typo in Falcon configuration options section. Should be image.repository

Issue #428, opened 2025-07-29
Falcon IAR documentation inconsistent repo flag name


dworkman1:
In the IAR helm deployment documentation I see "image.repo" mentioned in the table [Falcon configuration options](https://github.com/CrowdStrike/falcon-helm/tree/main/helm-charts/falcon-image-analyzer#falcon-configuration-options), but in the sample values file I see "image.repository". Is it intended to allow both?

PR #479, opened 2025-12-12
docs(sensor): update node sensor docs with allowed secret name for GKE autopilot

mr-jungchoi:
Updating falcon-sensor docs with the allowed `falconSecret.secretName`. All other secret names are disallowed.

This PR also includes doc updates to address the following issues:
- https://github.com/CrowdStrike/falcon-helm/issues/462
- https://github.com/CrowdStrike/falcon-helm/issues/428


helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -241,6 +241,18 @@ Validate one of falcon.cid or falconSecret is configured
 {{- end -}}
 {{- end -}}
 
+{{/*
+Validate falconSecret.secretName
+*/}}
+{{- define "falcon-sensor.validateFalconSecretName" -}}
+{{- $gkeAutopilotEnabled := (.Values.node.gke.autopilot . | eq "true") -}}
```
gpontejos-cs: Should we switch to dig here just in case someone doesn't have gke defined under node? 
```
{{- $gkeAutopilotEnabled := (dig "node" "gke" "autopilot" "false" .Values | eq "true") }}
```

gpontejos-cs (approved):

PR #500, opened 2026-02-19
chore(falcon-image-analyzer): update default image repository for unified image

cs-pvyas:
CSPG-81448 add unified image


helm-charts/falcon-image-analyzer/values.yaml
```diff
@@ -182,6 +185,7 @@ iarAgentService:
 kac:
   namespace : "falcon-kac"
 
+
```
mr-jungchoi: nit: unnecessary new line

PR #514, opened 2026-04-21
fix(falcon-kac): ensure watcher config values are handled properly

mr-jungchoi:


redhatrises (approved):

mr-jungchoi:
This change is actually not necessary for helm. The current helm behavior is correct.

PR #525, opened 2026-06-17
chore: add cid to falcon secrets for image analyzer chart

gpontejos-cs:


helm-charts/falcon-image-analyzer/values.schema.json
```diff
@@ -336,36 +336,6 @@
...
-                        "required": ["cid"],
-                        "properties": {
-                            "cid": {
-                                "type": "string"
-                            }
-                        }
-                    }
-                }
-            }
-        },
```
mr-jungchoi, line 368: Since we are removing this schema validation to catch when CID is missing for IAR, we should add a helper validation function. Here is an example for falcon-kac: https://github.com/CrowdStrike/falcon-helm/blob/main/helm-charts/falcon-kac/templates/_helpers.tpl#L196-L211

mr-jungchoi (changes requested):

helm-charts/falcon-image-analyzer/values.schema.json
gpontejos-cs: Re-added with some other validations.

mr-jungchoi (approved):

PR #527, opened 2026-06-25
chore: remove unused daemonset sensor permissions

gpontejos-cs:


pflanagan-cs (approved):
LGTM
Needs verify with end-to-end tests

helm-charts/falcon-sensor/templates/clusterrole.yaml
```diff
@@ -25,13 +25,6 @@ rules:
   - pods
   - services
   - nodes
-  - daemonsets
-  - replicasets
-  - deployments
-  - jobs
-  - ingresses
-  - cronjobs
-  - persistentvolumes
```
pflanagan-cs, line 34: IIUC these permissions are not using the correct `apiGroups` so would never have worked in any case

mr-jungchoi (approved):

PR #531, opened 2026-07-06
feat(falcon-sensor): add new falcon config volume for daemonset

mr-jungchoi:


helm-charts/falcon-sensor/templates/_helpers.tpl
```diff
@@ -145,7 +145,7 @@ args:
       else
       echo "Running /opt/CrowdStrike/configure-cluster-id";
       /opt/CrowdStrike/configure-cluster-id;
-      fi
+      fi;
```
gpontejos-cs: This semicolon isn't needed since it closes the final if statement.

gpontejos-cs (approved):

PR #538, opened 2026-07-28
chore: add HELM_CHART env to all sensor pods

mr-jungchoi:


gpontejos-cs (approved):

helm-charts/falcon-sensor/templates/tests/test-ds-helm-chart-env.yaml
```diff
@@ -0,0 +1,120 @@
+{{- if .Values.testing.enabled -}}
+{{- if .Values.node.enabled }}
```
gpontejos-cs, line 2: Let's also run this against the sidecar deployment

gpontejos-cs (changes requested):

gpontejos-cs (approved):

